PR #2, opened 2018-02-07
Add maven build

rleigh-codelibre:
- Move autogen sources to top level
- Move sphinx sources to top level
- Build sphinx docs with maven, including linkchecking in the test phase (`-DskipSphinxTests=true` or `-DskipTests=true` to skip)
- Build with travis and appveyor, (excluding linkchecking)
- Package for distribution

Testing: Check builds are green.  Build by hand if you like, since it will test the typical development workflow when doing a local build.  Generated documentation is zipped+tarred for distribution or deployment, and the build tree is in `target/sphinx/html`.

`mvn -DskipSphinxTests` will skip linkchecking.  Useful to skip selectively when built as part of a larger collection, where `-DskipTests` would skip every test, while `-DskipSphinxTests` is more selective.


joshmoore:
`The job exceeded the maximum time limit for jobs, and has been terminated.` during link checking.

rleigh-codelibre:
@joshmoore I've added the ability to skip the link checking.

joshmoore:
A big diff, so I didn't go through everything overly closely. Travis is green and the non-move changes make sense. Merging so we can iterate.

PR #3, opened 2018-02-20
Generate documentation with autogen 

rleigh-codelibre:
- Build autogen source
- Run generation steps
- Remove generated files from git

Testing:

Check all pages are present from the old bioformats docs.  Check the generated summary metadata and formats pages all look sane.

I've imported all the changes up to 5.8.0, so the content should be identical to current bioformats.git modulo some broken model links (because we don't have -SNAPSHOT versions staged).

PR #4, opened 2018-02-27
build: Allow skipSphinxTests to work with or without arguments

rleigh-codelibre:
Testing: Run https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge/ with or without validation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,7 +184,8 @@
       <id>sphinx-linkcheck</id>
       <activation>
         <property>
-          <name>!skipSphinxTests</name>
+          <name>skipSphinxTests</name>
+          <value>!true</value>
         </property>
       </activation>
       <build>


rleigh-codelibre:
@joshmoore Validation can be enabled now with this option.  Note that it will currently always fail when enabled due to some of the docs links being nonexistet for maven -SNAPSHOT versions.

joshmoore:
Thanks for this, @rleigh-codelibre. For those reading along, the parameter `SKIP_DOCS_VALIDATION` gets used under advanced maven settings as:

```
skipSphinxTests=${SKIP_DOCS_VALIDATION}
```

rleigh-codelibre:
This is also a more specialised form of the general `skipTests` parameter (which can also be used to skip everything, including the sphinx linkchecking).  `skipSphinxTests is specifically for skipping the sphinx linkcheck tests, rather than all tests.

joshmoore:
I've tested this locally from https://github.com/snoopycrimecop/bio-formats-build/tree/east_merge_trigger (using a Dockerfile that I'll describe under https://github.com/ome/bio-formats-test/pull/1). alicona is still failing for me locally, so I'd be inclined to change the default here to `true` though we'll still need a way to pass this value from the top level down to the submodules.

rleigh-codelibre:
@joshmoore The default can be changed, but that means not being in line with the existing maven conventions and options.  It's set appropriately for the documentation build in the job.  We can also create a makefile or other wrapper script which invokes maven with the appropriate options to match ome-documentation.

rleigh-codelibre:
Note: integrated with #3 including a `Makefile` to with separate `html` and `linkcheck` targets, mirroring the existing ome-model docs `Makefile` behaviour.

PR #5, opened 2018-02-28
Improve linkchecker

hflynn:
Implements https://github.com/openmicroscopy/ome-internal/pull/381/ to improve the linkchecking for the docs build


rleigh-codelibre:
Being built via maven, with a generated `conf.py`, we could pass the settings in as maven properties.  But the environment is certainly fine for now.

rleigh-codelibre:
Thanks, this change is good to merge.  @joshmoore please could you merge it?

I've tested locally and it works fine.  It failed one URL because sourceforce is down, but that's expected right now.

PR #12, opened 2018-03-29
build: Disable hard failure for linkchecking

rleigh-codelibre:
Maven won't allow this to be a configurable property (`-W`), so remove until we adopt gradle.  This will make the CI jobs succeed if the linkchecking fails.  It will require extra job logic to scan the logs for broken links.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -534,7 +534,6 @@
                     <argument>${sphinx_builddir}/cache</argument>
                     <argument>-b</argument>
                     <argument>${sphinx.linkchecker}</argument>
-                    <argument>-W</argument>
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
Tested here: https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge-docs/jdk=JDK9,label=testintegration/2/consoleFull

You can see there are multiple broken links and the build still failed.  This matches the existing job behaviour.

hflynn:
👍 console output looks as I'd expect (aside from I'm not sure why it's trying to point at a new model version but that's not really relevant to this PR)

Issue #13, opened 2018-03-30
Post decoupling content updates


sbesson:
See https://github.com/openmicroscopy/bioformats/pull/3117. As this repository is currently scheduled to become the new location for the Bio-Formats documentation from 5.8.2, content on a few pages will require some adjustments before moving to production:

- http://docs.openmicroscopy.org/bio-formats/5.8.1/developers/building-bioformats.html probably requires an update since the documentation will be built separately
- http://docs.openmicroscopy.org/bio-formats/5.8.1/developers/components.html will need updates to document the new decoupled components (examples and documentation)
- http://docs.openmicroscopy.org/bio-formats/5.8.1/developers/format-documentation.html will need a review since these pages are now auto-generated.
- http://docs.openmicroscopy.org/bio-formats/5.8.1/developers/useful-scripts.html could drop the `sphinx` mention

rleigh-codelibre:
PR #15 addresses points (2) and (4).

(1) currently does not include any examples or documentation-related material.  It can likely be revised to document bio-formats-build in the near future.

hflynn:
PR #16 updates 3. developers/format-documentation.html

PR #15, opened 2018-04-02
Post-decoupling changes

rleigh-codelibre:
Address some of the points in https://github.com/ome/bio-formats-documentation/issues/13

### DIFF
--- a/sphinx/developers/components.rst
+++ b/sphinx/developers/components.rst
@@ -171,8 +171,7 @@ Decoupled OME data model components:
 
 `Bio-Formats examples <https://github.com/ome/bio-formats-examples>`_
 
-Usage examples for using Bio-Formats with the Maven or Gradle build
-systems.
+Usage examples for Bio-Formats with Maven and Gradle.
 
 .. _bio-formats-documentation:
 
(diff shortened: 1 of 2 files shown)


sphinx/developers/components.rst
```diff
@@ -183,6 +167,19 @@ Decoupled OME data model components:
 - :ref:`OME-XML <ome-xml>`
 - :ref:`Specification <specification>`
 
+.. _bio-formats-examples:
+
+`Bio-Formats examples <https://github.com/ome/bio-formats-examples>`_
+
+Usage examples for using Bio-Formats with the Maven or Gradle build
```
hflynn: Minor nitpick but do we need both "usage" and "using" in this sentence? Just "Examples for using..." seems like it would be fine

sphinx/developers/components.rst
```diff
@@ -183,6 +167,19 @@ Decoupled OME data model components:
...
+systems.
+
+.. _bio-formats-documentation:
+
+`Bio-Formats documentation <https://github.com/ome/bio-formats-documentation>`_
+
+The Sphinx source repository for this manual.
```
hflynn: and HTML pages?

hflynn: Nevermind, I guess I was thinking of when we used to have more than one output

PR #18, opened 2018-04-06
Add ome_model.uri property to configure location of ome-model documentation

rleigh-codelibre:
Testing: will be tested on east-ci which will set the property.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,9 +33,12 @@
     <sphinx.ome_source.user>ome</sphinx.ome_source.user>
     <sphinx.openmicroscopy_source.user>openmicroscopy</sphinx.openmicroscopy_source.user>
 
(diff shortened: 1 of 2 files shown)


rleigh-codelibre:
See https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge-docs/jdk=JDK9,label=testintegration/19/console

The model links are "local", solving the immediate problem of the checking failures for snapshot model versions.  However, the ome-common, ome-xml and specification javadocs are missing.  We will need to solve these separately.

dgault:
Without this PR there were 225 broken links, mostly model related such as the example below:
`(line    9) broken    https://docs.openmicroscopy.org/ome-model/5.6.4-SNAPSHOT/ome-tiff`

With the PR included the broken ome-model links are cleaned up and there are only 25 remaining failures due to ome-common and ome-xml.

The addition of the ome_model.uri property is working as expected and the PR is good to merge.

PR #21, opened 2018-04-09
Bump component to 5.8.2-SNAPSHOT

sbesson:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
 
   <groupId>ome</groupId>
   <artifactId>bio-formats-documentation</artifactId>
-  <version>5.8.1</version>
+  <version>5.8.2-SNAPSHOT</version>
 
   <name>Bio-Formats documentation</name>
   <description>Bio-Formats Sphinx documentation</description>


rleigh-codelibre:
It's still pulling the 5.8.1 sources.  Is that correct, or should it be getting the 5.8.2-SNAPSHOT sources from ome.snapshot?

If it's correct, then should be fine to merge.

sbesson:
My understanding was that:

- like any other dependency, we do not want to bump `bioformats.version` to a SNAPSHOT version in a stable branch
- the consistent bump of all components to 5.8.2-SNAPSHOT is ensured by https://github.com/ome/bio-formats-build/blob/master/scripts/update-versions and pushed to the integration branch.

What would be the main benefits to bumping `bioformats.version` on `master`?

rleigh-codelibre:
It really depends upon whether we are planning to let the bioformats and bio-formats-documentation patch versions diverge, or whether we are planning to keep them in sync.  If the latter, then bumping both versions means that we're always building against the matched version of bioformats.  If the former, it doesn't matter quite so much, but still it does mean that it's not possible to document any additions until we're actually made a bioformats release, which isn't so nice.

sbesson:
Closing - superseded by https://github.com/ome/bio-formats-documentation/pull/24 (we might want to bump to 5.9.0-SNAPSHOT right after merging the former)

PR #28, opened 2018-05-16
Update public format page for Vectra QPTIFF format

sbesson:
See https://docs.openmicroscopy.org/bio-formats/5.8.2/formats/perkinelmer-vectra-qptiff.html

This updates the Vectra QPTIFF format page to:
- mention the pyramid support
- add references to the public specification and sample images

2028690 also removes a single `.gitignore` file in the legacy `docs/sphinx` folder (pre-decoupling).

/cc @melissalinkert

### DIFF
--- a/docs/sphinx/.gitignore
+++ b/docs/sphinx/.gitignore
@@ -1,3 +1,0 @@
-_build
-venv
-conf.py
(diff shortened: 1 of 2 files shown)


rleigh-codelibre:
The changes look fine.  However, I don't see the `pyramid` setting used in the output.

sbesson:
Should be staged at https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-docs/. I think the pyramid flag is only displayed in the supported formats table.

PR #30, opened 2018-06-20
Add license/copyright section to about page

melissalinkert:
This makes the license and current copyright easier to find, especially for anyone who uses Bio-Formats but doesn't read the code.

I'm not necessarily attached to the change in 968127a, so happy to revert if it's a problem or there is a better way to link to https://www.openmicroscopy.org/licensing/

### DIFF
--- a/sphinx/about/index.rst
+++ b/sphinx/about/index.rst
@@ -31,7 +31,9 @@ Copyright (C) 2005 - 2018 Open Microscopy Environment:
 
 Bio-Formats can be cited in publications as follows:
 
-`Melissa Linkert, Curtis T. Rueden, Chris Allan, Jean-Marie Burel, Will Moore, Andrew Patterson, Brian Loranger, Josh Moore, Carlos Neves, Donald MacDonald, Aleksandra Tarkowska, Caitlin Sticco, Emma Hill, Mike Rossner, Kevin W. Eliceiri, and Jason R. Swedlow (2010) Metadata matters: access to image data in the real world. The Journal of Cell Biology 189(5), 777-782. doi: 10.1083/jcb.201004104 <http://www.ncbi.nlm.nih.gov/pubmed/20513764>`_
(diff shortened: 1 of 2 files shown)


sbesson:
💯 for promoting this information. I'll review the content in more details tomorrow.
One related RFE: could we  include a paragraph about citing Bio-Formats alongside the license information i.e. referring to Linkert et al 2010?

sphinx/about/index.rst
```diff
@@ -18,6 +18,23 @@ structure is of vital importance to the community. You may find LOCI's article
...
+Unless otherwise noted, the current copyright statement is:
+
+Copyright (C) 2005 - 2018 Open Microscopy Environment:
+      - Board of Regents of the University of Wisconsin-Madison
+      - Glencoe Software, Inc.
+      - University of Dundee
+
+Bio-Formats can be cited in publications as follows:
+
+`Melissa Linkert, Curtis T. Rueden, Chris Allan, Jean-Marie Burel, Will Moore, Andrew Patterson, Brian Loranger, Josh Moore, Carlos Neves, Donald MacDonald, Aleksandra Tarkowska, Caitlin Sticco, Emma Hill, Mike Rossner, Kevin W. Eliceiri, and Jason R. Swedlow (2010) Metadata matters: access to image data in the real world. The Journal of Cell Biology 189(5), 777-782. doi: 10.1083/jcb.201004104 <http://www.ncbi.nlm.nih.gov/pubmed/20513764>`_
```
sbesson: Only concern is that this makes the hyperlink quite long. I would be inclined to use the same formatting as in https://www.openmicroscopy.org/citing-ome/ and have an extra line with the PubMed ID linking to the paper.

melissalinkert: cb1a217 should fix this

sphinx/about/index.rst
```diff
@@ -18,6 +18,25 @@ structure is of vital importance to the community. You may find LOCI's article
 on `open source software in
 science <http://loci.wisc.edu/software/oss>`_ of interest.
 
+License, copyright, and citation
+--------------------------------
+
+Licensing information for Bio-Formats is described on :oo_root:`the OME licensing page <licensing>`.
+Unless otherwise noted, the current copyright statement is:
```
mtbc: If we say this is current, is there a process for keeping this page up to date?

PR #39, opened 2018-07-04
Update link to i3dcore

melissalinkert:
See https://trello.com/c/lvf099f3/234-i3dcore-cbia-3d-library-status.  i3d seems to have its own page on the CBIA website again, so it's now linked directly.


sbesson:
Generally the link to this 3d party is still ignored via https://github.com/ome/bio-formats-documentation/blob/v5.9.0/sphinx/conf.py#L389. Is it worth keeping it ignored or should we try to re-include?

melissalinkert:
2274a68 removes the exclude.  The linkcheck build seems fine locally, but can revert if it causes a problem.

sbesson:
Links are still failing with invalid SSL certs. Pushed cfb6e3b reverting the link re-inclusion.

PR #37, opened 2018-07-03
Bump version to 5.9.1-SNAPSHOT

dgault:
Bumping version to 5.9.1-SNAPSHOT

PR #40, opened 2018-07-10
Bump version to 5.9.1-SNAPHSOT

sbesson:


sbesson:
@dgault: this should replace #37. Vetoes to merging (before we start integrating the existing doc PRs)

PR #42, opened 2018-07-31
Update the link to www.slidebook.com

sbesson:
The SSL certificate has expired (causing the linkcheck to fail) and this page is moved permanently to the 3i website.


sbesson:
https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-docs/ now green. Proposing to merge unless there is a better alternative link /cc @dgault @melissalinkert

melissalinkert:
:+1:

PR #43, opened 2018-08-07
Permanent redirects

sbesson:
See https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-docs/134/. One of the URLs broke the linkcheck with `Exceeded 30 redirects.`

This PR reviews some of the URLS marked as `Permanently redirected` by the linkchecked, primarily in the components and format pages.

- use GitHub `sourcedir` extlink rather than `source` for components
- fix the organization of the https://github.com/ome/bio-formats-examples repository
- fix most of the permanently redirected URLs in the Format pages, switching to HTTPS whenever possible


dgault:
Changes look good and the docs builds have been green with this included

PR #49, opened 2018-08-30
Update public samples in format-pages.txt

sbesson:
We have public samples for several formats under https://downloads.openmicroscopy.org/images/ that are not cross-referenced from the technical format pages.

To test this PR, https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-docs/ should remain green and the generated format pages should contain hyperlinks to the public sample folders

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -410,7 +410,8 @@ weHave = * an `Imaris (RAW) specification document <http://flash.bitplane.com/wd
 * Bitplane's bfFileReaderImaris3N code (from no later than 2005, in C++) \n
 * several older Imaris (RAW) datasets \n
 * one Imaris 3 (TIFF) dataset \n
-* several Imaris 5.5 (HDF) datasets
+* several Imaris 5.5 (HDF) datasets\n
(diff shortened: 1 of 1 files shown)


mtbc:
`extensions = .mrc` yet https://downloads.openmicroscopy.org/images/MRC/ just has `EMD-2225.map`

mtbc:
`extensions = .sxm` yet https://downloads.openmicroscopy.org/images/LEO/ just has `2160grating*.tif`

mtbc:
Apart from those couple of puzzlements which may be outwith this PR all looks good.

sbesson:
@mtbc pushed another commit to update the relevant extensions to match what is auto-generated in https://docs.openmicroscopy.org/bio-formats/5.9.1/formats/dataset-table.html.

More generally, happy to hear thoughts how we make sure this metadata is kept up-to-date and/or consolidate the two tables into one place only. The main challenge is that the mapping between what we call formats and datasets (i.e. Bio-Formats readers) is not strictly one-to-one. A typical example is the Imaris IMS which is represented as one format page but three readers.

PR #53, opened 2018-10-01
Replace www.farsight-toolkit.org by farsight-toolkit.ee.uh.edu

sbesson:
The org domain seems to have gone away. This updates the documentation
to use another domain which serves the same content


dgault:
All the updated links match he content of the original and are correct and working

PR #54, opened 2018-10-01
Fix Javadoc extlinks to use latest for non-release versions

sbesson:
- uses the https://javadoc.io/page base URL as recommended upstream
- conditionally use latest Javadoc URL when the component version ends up in SNAPSHOT

With this and #53 https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge-docs/ should turn green.


dgault:
Using latest for the SNAPSHOT version makes sense and looks good. The build failures in https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge-docs/jdk=JDK9,label=testintegration/ are related to https://www.slf4j.org which has been temperamental through out the day so far. Build should be green again shortly as that link seems to be working much better this afternoon

PR #56, opened 2018-10-10
 matlab-dev.rst sphinx\developers

kouichi-c-nakamura:
More details about Jave heap memory setting for MATLAB

I edited the same section before, but I realized that the maximum value for Java heap memory size allowed in MATLAB Preferences GUI is limited. If you need more size, you need to overwrite the value by editing the `matlab.prf` in the directory given by `prefdir` MATLAB function, or the `java.opts` in the `matlabroot/bin/arch` folder.

https://uk.mathworks.com/matlabcentral/answers/337797-java-heap-size-is-not-accessing-all-available-ram


sphinx/developers/matlab-dev.rst
```diff
@@ -23,13 +23,38 @@ MATLAB can be retrieved using::
...
+::
+
+  JavaMemHeapMax=I8800
+
+java.opts
+^^^^^^^^^
+
+Alternatively, this can also be done by creating a :file:`java.opts` file in
+`the startup directory <https://mathworks.com/help/matlab/matlab_env/matlab-startup-folder.html>`_
+ and overriding the default memory settings (see `this page <https://mathworks.com/help/matlab/matlab_env/java-opts-file.html>`_
```
sbesson: The leading space causes the Travis check to fail - see https://travis-ci.org/ome/bio-formats-documentation/jobs/439770938#L1573

```
Warning, treated as error:
/home/travis/build/ome/bio-formats-documentation/target/sphinx/developers/matlab-dev.rst:54:Unexpected indentation.
```

kouichi-c-nakamura:
Thanks for testing. I committed a revised version.

sphinx/developers/matlab-dev.rst
```diff
@@ -23,13 +23,37 @@ MATLAB can be retrieved using::
...
+MATLAB Preferences (R2010a+)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Alternatively this can also be done by creating a :file:`java.opts` file in
-the startup directory and overriding the default memory settings. We recommend
-using ``-Xmx512m`` in your :file:`java.opts` file. Calling::
+Default |JVM| settings can be increased by `Java Heap Memory Preferences
+<https://mathworks.com/help/matlab/matlab_external/java-heap-memory-preferences.html>`_
+of MATLAB (R2010a onwards) using
+:menuselection:**`Preferences --> General --> Java Heap Memory`**.
```
sbesson: The `**` need to be removed as they interfer with the markup.

sphinx/developers/matlab-dev.rst
```diff
@@ -23,13 +23,37 @@ MATLAB can be retrieved using::
...
+
+.. image:: ../images/matlab_memory_pref.png
+
+However, the maximum value allowed in the Preferences GUI can still be too small
+for your purpose. In that case, you can directly edit :file:``matlab.prf`` file in the
```
sbesson: Single quotes need to be used with the `file` markup.

sphinx/developers/matlab-dev.rst
```diff
@@ -23,13 +23,37 @@ MATLAB can be retrieved using::
...
+
+  JavaMemHeapMax=I25000
+
+java.opts
+^^^^^^^^^
+
+Alternatively, this can also be done by creating a :file:`java.opts` file in
+`the startup directory <https://mathworks.com/help/matlab/matlab_env/matlab-startup-folder.html>`_
+and overriding the default memory settings (see `this page <https://mathworks.com/help/matlab/matlab_env/java-opts-file.html>`_
+for more informaiton). We recommend using ``-Xmx512m`` in your :file:`java.opts`
```
sbesson: information

sphinx/developers/matlab-dev.rst
```diff
@@ -23,13 +23,37 @@ MATLAB can be retrieved using::
...
+<https://mathworks.com/help/matlab/matlab_external/java-heap-memory-preferences.html>`_
+of MATLAB (R2010a onwards) using
+:menuselection:**`Preferences --> General --> Java Heap Memory`**.
+
+.. image:: ../images/matlab_memory_pref.png
+
+However, the maximum value allowed in the Preferences GUI can still be too small
+for your purpose. In that case, you can directly edit :file:``matlab.prf`` file in the
+folder specified by the ``prefdir`` `MATLAB function <https://mathworks.com/help/matlab/ref/prefdir.html>`_
+(eg. ``'C:\Users\xxxxxxx\AppData\Roaming\MathWorks\MATLAB\R2018b'``). Find the
```
sbesson: The `file` markup can probably be used here as well

kouichi-c-nakamura:
> The only concern I have from the content is that the the -Xmx512m
recommendation got moved to the second section while the first section 
uses a screenshot and example suggesting 25G. Could this be interpreted 
as confusing by consumers i.e. do we need either a top-level statement 
indicating we recommend setting the memory to 512M or updating the 
screenshot and snippet to use the recommended (base) value?

Thank for the detailed review and comments. 

Actually, I thought about what to show in the screenshot. In my environment, the default value set by MathWorks is 768 MB (this value might be fixed), so setting the heap size to 512 MB, i.e. decreasing from the default value, does not make good sense. The maximum allowed by default in my environment is 8,168 MB (this value appears to be environment-dependent; in another machine it says 2,348 MB).

Do you recommend 512 MB for a reason? If not, how about using 1024 MB as a common recommendation using Preferences with a screenshot and `java.opts`? Then we can add a subsection for the Preferences about how to increase the heap size over the default limit with another screenshot, eg. 25GB.

sbesson:
@kouichi-c-nakamura I think the rationale for using 512m is that it aligns with what is set in the command-line tools - see https://github.com/openmicroscopy/bioformats/blob/v5.9.2/tools/bf.sh#L27.
I think this value was identified as a good default/mininum value as it proved to work with a most file formats by default. I agree that if the default JVM settings are greater than this value, there is no reason to decrease it.

kouichi-c-nakamura:
Thanks. So the choices are:

1. Stick to the good old 512MB.
2. Change the values to 768 MB

If we choose 2, do we need to change the values for command-line tools as well? Probably not because they are not the same tools.

sbesson:
Hi Kouichi, sorry for the delay. Reading various resources, it looks like the default allocated heap space might correspond to 25% of the memory so the default value will be system-dependent. I do not think upgrading the minimum value at this stage is worth the effort especially as the required memory will depend on the file format and the data that is being read.

In terms of action, I am happy to limit the scope of this PR to fixing the formatting comments reported above  in order to get it merged.

kouichi-c-nakamura:
Understood. We can actually mention that the default heap size is system-dependent, while keeping the good old 512M as our example.

Anyway, I'll soon commit a reasonable version. Tell me what you think then. Thanks.

sbesson:
screen shot 2018-11-02 at 11 32 42

Looks good. Thanks @kouichi-c-nakamura. Merging and we can go through more iterations on this page as separate PRs if needed.

kouichi-c-nakamura:
Thanks a lot for your review.

PR #58, opened 2018-11-13
 xml-validation.rst sphinx\users\comlinetools

kouichi-c-nakamura:
In this PR, I edited files under `sphinx\users\cmlinetools`

# 1. Go back link
Added 
```
Go back to :doc:`index`
```
at the bottom of the pages. I thought it was difficult to navigate without this link.

# 2 xmlvalid

Added example of use of `xmlvalid`

Fixed a bug in an example of use of `xmlvalid` and `tiffcomment` (removed a hyphen)

See also a related discussion in the forum:
https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8620&p=20225#p20225

### DIFF
--- a/sphinx/users/comlinetools/conversion.rst
+++ b/sphinx/users/comlinetools/conversion.rst
@@ -169,5 +169,3 @@ name pattern, then the other must be included too.  The only exception is if
       bfconvert /path/to/input output_xy%sz%zc%ct%t.ome.tif -padded
 
     .. versionadded:: 5.2.2
-
-Go back to :doc:`index`.
--- a/sphinx/users/comlinetools/display.rst
+++ b/sphinx/users/comlinetools/display.rst
@@ -172,5 +172,3 @@ dimensions, and other basic metadata will be printed to the console.
     will be opened with ``loci.formats.in.ImageReader``::
 
       showinf -format APNG test.png
-
-Go back to :doc:`index`.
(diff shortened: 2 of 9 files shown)


sbesson:
Thanks @kouichi-c-nakamura. Labelling for inclusion in the CI builds so that we can review in the upcoming days.

One question about the go back link. In https://docs.openmicroscopy.org/bio-formats/5.9.2/index.html or any page, this is achieved by the `Bio-Formats 5.9.2 documentation` hyperlink at the top-left and bottom-left of every generated HTML page. Happy to envision solutions to improve the navigation although I am concerned about the maintenance burden of adding a text link to every source page. Can this change be moved either as a separate commit  or even as a separate PR as it is effectively independent of the utility issue reported in the forum thread?

kouichi-c-nakamura:
Aha, to tell the truth, I didn't notice those links 

But if I take an example of [this page](https://docs.openmicroscopy.org/bio-formats/5.9.2/users/comlinetools/edit.html), this is a page about Command Line Tool, so I'd love to have a link to [the top page of Command Line Tool](https://docs.openmicroscopy.org/bio-formats/5.9.2/users/comlinetools/index.html). However, the links at the top or the bottom will take you to [one level further up](https://docs.openmicroscopy.org/bio-formats/5.9.2/users/index.html).

But it's really a small thing. If you don't find it useful, feel free to ignore :)

sbesson:
Ah I see what you expect now. I can definitely conceive having another level in the top-level navigation bar for the command-line tools as we are promoting these as a subsection of the Users page.

At the moment, the `users/index.rst` page creates a flat hierarchy of subpages. One option would be to create a CL tools hierarchy using `users/comlinetools/index.rst`. I have pushed an example of doing so at https://github.com/sbesson/bio-formats-documentation/tree/cl_index. Please have a look and let me know if that would be a solution.

kouichi-c-nakamura:
I like your idea. It's a smarter solution.

Cause I don't know how to remove the changes I made in the commit history, I'm going to make a commit to undo the 'go back' changes. Is it OK?

sbesson:
Yes, partially reverting 9459d41 is fine. In general, we encourage contributors to keep commits atomic i.e. dedicated to one set of changes as standard best practice.

kouichi-c-nakamura:
Thanks. A new commit has been made.

Yes, I agree that. Just a bit I wasn't careful this time.

kouichi-c-nakamura:
Thanks as always!

PR #60, opened 2018-11-17
Users documentation: create a new level of hierarchy for the command-line tools

sbesson:
Opening  for discussion following the thread started in #58 

This PR proposes to refactor the Sphinx toctree hierarchy for the `users` section of the technical documentation to create an extra level for the command-line tools.

This already matches the organization of the rST source files under a single `comlinetools/` folder. The main advantage of this structure so is that it creates an extra layer of hierarchy in the top-level navigation bar allowing to go back to the Command line tools index page rather than back to the Users index page.

### DIFF
--- a/sphinx/users/comlinetools/index.rst
+++ b/sphinx/users/comlinetools/index.rst
@@ -1,8 +1,22 @@
-Command line tools introduction
-===============================
+Command line tools
+==================
 
-There are several scripts for using Bio-Formats on the command
-line.
+There are several scripts for using Bio-Formats on the command line.
+
+.. toctree::
+    :maxdepth: 1
+    :titlesonly:
+    :hidden:
+
+    display
+    conversion
+    xml-validation
+    edit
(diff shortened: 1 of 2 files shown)


joshmoore:
Any further comments @kouichi-c-nakamura?

sbesson:
Yes the absence of hierarchy on the index page containing the `toctree` is slightly disconcerting at first. This is already the case for other pages  e.g. in https://docs.openmicroscopy.org/bio-formats/5.9.2/users/index.html (no `User Information` hierarchy in the top-level breadcrumbs). I have not checked whether there is a Sphinx style/option allowing to modify this.

kouichi-c-nakamura:
I see what @sbesson means. If you can find a way, that should be less confusing for sure.

Thank you @joshmoore for accepting my proposal. I think this is better!

kouichi-c-nakamura:
Thanks!

PR #61, opened 2018-11-21
Use HTTPS for Nikon external URLS

sbesson:
This should address https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge-docs/jdk=JDK8,label=testintegration/250/

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -404,4 +404,6 @@ linkcheck_ignore = ['https://imspector.mpibpc.mpg.de',
     'http://cellularimaging.perkinelmer.com/downloads/',
     'https://animatedpngs.com/', # SSL certificate error
     'https://www.merckmillipore.com', # Read timeout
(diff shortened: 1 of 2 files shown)


sbesson:
Still returning 

```
[sbesson@ome-web-prod ~]$ curl -I https://www.nikoninstruments.com
curl: (60) Peer certificate cannot be authenticated with known CA certificates
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
```

Merging and we can review later if the SSL cert gets fixed upstream.

PR #65, opened 2018-12-07
Direct user feedback to image.sc

joshmoore:
In preparation for I2K presentation.

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -123,8 +123,11 @@ oo_root = 'https://www.openmicroscopy.org'
 lists_root = 'http://lists.openmicroscopy.org.uk'
 downloads_root = 'https://downloads.openmicroscopy.org'
 docs_root = 'https://docs.openmicroscopy.org'
+imagesc_root = 'https://forum.image.sc'
 
 extlinks = {
+    # image.sc
+    'imagesc' : (imagesc_root + '/%s', '#'),
     # Trac links
     'ticket' : (trac_root + '/ticket/%s', '#'),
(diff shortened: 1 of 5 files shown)


sphinx/conf.py
```diff
@@ -123,8 +123,11 @@
 lists_root = 'http://lists.openmicroscopy.org.uk'
 downloads_root = 'https://downloads.openmicroscopy.org'
 docs_root = 'https://docs.openmicroscopy.org'
+imagesc_root = 'https://forum.image.sc'
```
sbesson: Thoughts about pointing directly at https://forum.image.sc/tags/bio-formats in this technical documentation

mtbc:
With `make html` I get,

```
Warning, treated as error:
/home/mtbc/src/bio-formats-documentation/target/sphinx/about/index.rst:56: WARNING: Mismatch: both interpreted text role prefix and reference suffix.
```

joshmoore:
- @mtbc : hmmm, ... I had that before, but then I thought I fixed it. Perhaps it got lost.
 - @sbesson : can do, I'll update and we can compare.

joshmoore:
Pushed.

sbesson:
Overall looks good and matches the discussions and announcements from the I2K conference.  https://github.com/ome/bio-formats-documentation/blob/v6.0.0-m3/sphinx/about/bug-reporting.rst#sending-a-bug-report is another place which could be updated with the image.sc forum. Otherwise looks ready to get in.
 Outwith the scope of this PR is the question of whether we want to review the developer-targetted `pom.xml` and reference the forum.

PR #69, opened 2019-01-14
Fix link to View5D plugin

sbesson:
Previous link seemed to have ceased to exist

This should fix https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge-docs /cc @dgault


sbesson:
https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge-docs/321/

PR #72, opened 2019-01-28
Fix Becker & Hickl URLs

sbesson:
Both the specification and the SPCImage URLs seem to have been broken
following a reorganization of the manufacturer website.


sbesson:
https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge-docs/321/ is green with this change.

PR #74, opened 2019-02-04
Improve versioned links for Bio-Formats downloads

sbesson:
This primarily updates the logic pointing at the Bio-FOrmats downloads which were previously linking to the 5.8 downloads.

In addition, it refactors the logic to minimize the maintenance of these redirects for each major/minor version. 

- `bioformats.version` is now read in `conf.py`
- if the version is SNAPSHOT as in coupled daily builds, a latest URL is constructed using the `major.minor` version
- if the version is a release, the versioned URLs are used
- the changes above allowed to detect a broken link. eff14ce updates the links to APIs refactored in the `IPyramidHandler` interface


sbesson:
Merging as this has been in the CI docs builds for the last few days and no obvious issue has been noticed. @dgault has reported some issues while testing this locally. So I propose to get this in for 6.0.0-rc1 and retest the local build as well as the rc1 documentation to make sure all links to the source code/downloads/API are correct.

PR #75, opened 2019-02-06
Add documentation for GE MicroCT format

melissalinkert:
The reader was introduced in https://github.com/openmicroscopy/bioformats/pull/3227.

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -728,6 +728,18 @@ notes = Applications that support .dm3 files include `Datasqueeze <https://www.s
 \n
 Note that the Gatan Reader does not currently support stacks.
 
+[GE MicroCT]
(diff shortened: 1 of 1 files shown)


src/main/resources/format-pages.txt
```diff
@@ -728,6 +728,18 @@ notes = Applications that support .dm3 files include `Datasqueeze <https://www.s
 \n
 Note that the Gatan Reader does not currently support stacks.
 
+[GE MicroCT]
+extensions = .vff
+developer = `GE <http://gelifesciences.com/>`_
```
mtbc: could use `https`

melissalinkert: I had copied this from one of the other GE formats; b1d59d3 changes all GE links to use ```https``` for consistency.

PR #78, opened 2019-02-08
Add OME-TIFF pyramid section to WSI page

melissalinkert:
See https://trello.com/c/BX0CpQEM/65-documentation-page-to-link-to-specification-and-pyramid-writing-examples

PR #79, opened 2019-02-14
Add -noflat option to the showinf and bfconvert users pages

sbesson:
Noticed while reviewing #78. We currently do not mention the `-noflat` option either in the `showinf` page or in the `bfconvert` page (new 6.0.0 feature).

Not a blocker for 6.0.0 but it should be straightforward addition.

### DIFF
--- a/sphinx/users/comlinetools/conversion.rst
+++ b/sphinx/users/comlinetools/conversion.rst
@@ -32,9 +32,11 @@ The output file format is determined by the extension of the output file, e.g.
 .. option:: -noflat
 
     Do not flatten the series i.e. use the sub-resolution API for images
-    with pyramidal levels::
+    with pyramidal levels. This option is mandatory to generate an output
+    image with pyramidal levels. As of Bio-Formats 6.0.0, only the OME-TIFF
+    writer properly supports this option::
 
-      bfconvert -noflat /path/to/input output-first-series.tiff
+      bfconvert -noflat /path/to/input output-first-series.ome.tiff
 
     .. versionadded:: 6.0.0
 
--- a/sphinx/users/comlinetools/conversion.rst
+++ b/sphinx/users/comlinetools/conversion.rst
@@ -31,10 +31,10 @@ The output file format is determined by the extension of the output file, e.g.
 
 .. option:: -noflat
 
-    Do not flatten the series i.e. use the sub-resolution API for images
-    with pyramidal levels. This option is mandatory to generate an output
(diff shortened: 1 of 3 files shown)


sphinx/users/comlinetools/conversion.rst
```diff
@@ -29,6 +29,15 @@ The output file format is determined by the extension of the output file, e.g.
 
     .. versionadded:: 5.4.0
 
+.. option:: -noflat
+
+    Do not flatten the series i.e. use the sub-resolution API for images
+    with pyramidal levels::
```
melissalinkert: Worth mentioning that (1) this only matters if the output format is OME-TIFF, and (2) a pyramid will not be generated without it?

This PR also resolves https://trello.com/c/whGDjGxc/316-add-noflat-to-http-docsopenmicroscopyorg-bio-formats-592-users-comlinetools-displayhtml from the backlog.

rleigh-codelibre: With the new major version, wouldn't it be an opportune time to make `noflat` the default and to add `flat` options to enable flattening if desired?  This would make pyramid support the default.  We've left `flat` the default for a good six years now; isn't it time to change that?

sbesson: See f2c0cb5 for the 2 comments.

I definitely approve the idea of enabled sub-resolution by default especially now that Bio-Formats 6.0.0 has support for sub-resolution writing API and a concrete format implementing this API. I'll record this. Potentially the biggest thing to evaluate would be the impact when someone is converting to another format than `.ome.tiff` as it can lead to data loss.

sphinx/users/comlinetools/conversion.rst
```diff
@@ -29,6 +29,17 @@ The output file format is determined by the extension of the output file, e.g.
 
     .. versionadded:: 5.4.0
 
+.. option:: -noflat
+
+    Do not flatten the series i.e. use the sub-resolution API for images
```
mtbc: Perhaps could reword somehow: it's not clear if the "i.e. use the ..." explains the "do not X ..." or the "X ...": it's ambiguous if the "do not" distributes over it.

dgault: Agreed that it could be misinterpreted and would be best if reworded to remove the "ie" and make it clear if the sub-resolution API is being used or not .

dgault:
Other than the one comment this looks good and everything builds and displays as expected.

dgault:
Updated description looks good and reads well.

PR #84, opened 2019-02-26
Ignore Farsight toolkit links

sbesson:
These links have been reported as broken for the last builds


dgault:
Links still appear to be broken. Merging to ignore

PR #85, opened 2019-02-27
Remove link to large zip with MicroCT datasets

melissalinkert:
As discussed with @ome/formats, linking to a large file seems to slow down the docs build.  This removes the direct link to the problematic Zip, but keeps the link to the landing page.

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -732,7 +732,7 @@ Note that the Gatan Reader does not currently support stacks.
 extensions = .vff
 developer = `GE <https://gelifesciences.com/>`_
 bsd = no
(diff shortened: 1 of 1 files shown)


joshmoore:
No objections, but the fact that zips are being downloaded is otherwise worrisome.

sbesson:
Build time has certainly been reduced with this change but I agree with the concerns expressed in https://github.com/ome/bio-formats-documentation/pull/85#issuecomment-467898659. Looking at the linkcheck core logic, it looks like it should have been calling `requests.head` but I cannot reproduce the issue. I will try and create a minimal example to reproduce.

dgault:
Build times have reduced and the MicroCT dataset is still easy to locate from the link to the main archive page.

PR #95, opened 2019-04-21
Remove links to MATLAB Answers for using non-default JVM

sbesson:
These public threads seem to have been retired by Mathworks. The note is
updated to point to the public forum.

### DIFF
--- a/sphinx/users/matlab/index.rst
+++ b/sphinx/users/matlab/index.rst
@@ -15,7 +15,7 @@ image files. Note the minimum recommended MATLAB version is R2017b.
    It is possible to run Bio-Formats 6 on earlier MATLAB versions using a
    JVM version 8 or greater although using a different JVM than the one shipped
    with MATLAB can affect other functionalities. Please refer to the
-   `MATLAB Answers <https://uk.mathworks.com/matlabcentral/answers/index>`__
+   `MATLAB Answers <https://mathworks.com/matlabcentral/answers/index>`__
    for more information.
 
 Installation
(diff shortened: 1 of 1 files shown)


sphinx/users/matlab/index.rst
```diff
@@ -13,11 +13,10 @@ image files. Note the minimum recommended MATLAB version is R2017b.
...
 
    It is possible to run Bio-Formats 6 on earlier MATLAB versions using a
-   JVM version 8 or greater. Please refer to the support threads explaining
-   how to change the JVM software used by MATLAB for
-   `Mac OS <https://uk.mathworks.com/matlabcentral/answers/103056-how-do-i-change-the-java-virtual-machine-jvm-that-matlab-is-using-on-macos>`__,
-   `Linux <https://uk.mathworks.com/matlabcentral/answers/130360-how-do-i-change-the-java-virtual-machine-jvm-that-matlab-is-using-for-linux>`__
-   or `Windows <https://uk.mathworks.com/matlabcentral/answers/130359-how-do-i-change-the-java-virtual-machine-jvm-that-matlab-is-using-on-windows>`__.
+   JVM version 8 or greater although using a different JVM than the one shipped
+   with MATLAB can affect other functionalities. Please refer to the
+   `MATLAB Answers <https://uk.mathworks.com/matlabcentral/answers/index>`__
```
mtbc: Maybe drop the `uk.` to allow geo-located redirection.

mtbc:
Pity we can't give a more specific pointer but I appreciate what a pain MathWorks' site seems to be in this regard so this might be about as well as we can do.

sbesson:
https://uk.mathworks.com/matlabcentral/answers/99318-can-i-run-matlab-7-11-r2010b-with-jvm-1-5-older-version-by-using-the-matlab_java-environment-var#answer_108665 was the best alternate thread I was about to find. However suspect there has been some active culling of all threads where the `MATLAB_JAVA` recommendation was suggested by staff members and this one might be the next to go.

PR #96, opened 2019-04-22
Document other dependency updates included in 6.1.0-m1

sbesson:
Similarly to #92, this documents the dependency bumps performed in https://github.com/openmicroscopy/bioformats/pull/3252 https://github.com/openmicroscopy/bioformats/pull/3338 and https://github.com/openmicroscopy/bioformats/pull/3335

Note the file-leak dependency added in https://github.com/openmicroscopy/bioformats/pull/3330 is not listed as it is only available within the scope of the `test-suite` component.


mtbc:
Staged at https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge-docs/jdk=JDK8,label=testintegration/ws/bio-formats-build/bio-formats-documentation/target/sphinx/html/developers/java-library.html

PR #97, opened 2019-05-14
Temporarily ignore dicom.nema.org

sbesson:
See https://groups.google.com/forum/#!topic/comp.protocols.dicom/n4WXaSoafYU


sbesson:
The website seems to be back up and running. Closing.

PR #99, opened 2019-05-16
Bump version to 6.1.0

dgault:
Bump BF version to 6.1.0

PR #100, opened 2019-05-16
Add new format page for BDV

dgault:
New format page for Big Data Viewer

Ratings are based on https://docs.openmicroscopy.org/bio-formats/5.9.2/supported-formats.html#term-ratings-legend-and-definitions

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -338,7 +338,7 @@ mif = true
 
 [Big Data Viewer]
 extensions = .xml, .h5
-owner = `Tobias Pietzsch <https://imagej.net/BigDataViewer/>`_
+owner = `Tobias Pietzsch <https://imagej.net/BigDataViewer>`_
 bsd = no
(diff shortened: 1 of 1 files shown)


sbesson:
NB: this depends on https://github.com/ome/bio-formats-documentation/pull/99 as the reader points the BDVReader metadata page generated from the reader source code.

src/main/resources/format-pages.txt
```diff
@@ -336,6 +336,21 @@ privateSpecification = true
 reader = SDTReader
 mif = true
 
+[Big Data Viewer]
+extensions = .xml, .h5
+owner = `Tobias Pietzsch <https://imagej.net/BigDataViewer/>`_
```
sbesson: Looks like https://imagej.net/BigDataViewer/ is leading us nowhere and https://imagej.net/BigDataViewer should be used instead

sbesson:
Barring the issue about the URL, the rendered format page looks good.

PR #105, opened 2019-06-07
Add version history for v6.1.1

dgault:
For a full list of included PR's see https://github.com/openmicroscopy/bioformats/milestone/69?closed=1


sbesson:
https://github.com/openmicroscopy/bioformats/pull/3373 should also be included (just added the milestone) here and in https://github.com/ome/www.openmicroscopy.org/pull/299. Sean might need to be added to the external contributors page.

PR #110, opened 2019-06-26
Dcraw - fix broken link

dgault:
The original link to Dcraw software is throwing 404 Client Error: Not Found for url
This is due to the software being discontinued. As per the Formats meeting it was agreed to switch to using the Wikipedia entry for the software instead


dgault:
Also added an additional fox for a broken link to the downloader for Zeiss LSM Image Browser. Now just linking to the Zeiss downloads page instead unless anyone knows a better link to use.

PR #116, opened 2019-07-19
Fix URL to Roche Digital Diagnostics home page

sbesson:


jburel:
Without the change, it leads to a 404
With the change it works. 
Merging

PR #121, opened 2019-08-13
Use imagej.net link for LSM toolbox

sbesson:
A reference on the previous link is currently found on the image.net wiki entry.


dgault:
New link is correct and working. Good to merge.

PR #126, opened 2019-09-10
Add forward-compatible Python 3 support

sbesson:
To be tested via https://py3-ci.openmicroscopy.org/jenkins/


sbesson:
Merging as per https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-build-docs/

PR #128, opened 2019-09-23
Update Deltavision page to reference RCPNL

melissalinkert:
See https://github.com/ome/bioformats/pull/3421.

This won't build until ```bioformats.version``` is updated to a version that includes ```RCPNLReader```, so feel free to close until 6.3.0 if that's easier.


src/main/resources/format-pages.txt
```diff
@@ -582,6 +582,7 @@ notes = - The Deltavision format is based on the Medical Research Council (MRC)
   * `Bitplane Imaris <http://www.bitplane.com/>`_ \n
   * `SVI Huygens <http://svi.nl/>`_ \n
   * `Image-Pro Plus <http://www.mediacy.com/>`_ \n
+- RCPNL is a variant of the Deltavision format and requires a separate reader. \n
```
mtbc:
```suggestion
- RCPNL is a variant of the DeltaVision format and requires a separate reader. \n
```

PR #129, opened 2019-09-26
Document all known JPEG2000 suffixes

sbesson:
Noticed while working on https://github.com/ome/bioformats/pull/3445

Our format pages only indicate a subset of the file extensions that would be detected as JPEG2000 files.


dgault:
The updated list fo extensions here matches those used by the `JPEG2000Reader`. Good to merge

PR #131, opened 2019-10-11
Fix broken Perkin-Elmer links

dgault:
Fixes for broken links. Note the Openlab RAW specification link is removed altogether as I have been unable to find a link to the tech paper since they updated the website (though there does still seem to be a lot of broken links and missing content on their site at the moment and the search is not returning anything).

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -432,5 +432,6 @@ linkcheck_ignore = ['https://imspector.mpibpc.mpg.de',
     r'https://www.nis-elements.com/.*', # Invalid SSL certificate
     r'https://www.nikoninstruments.com/.*', # Invalid SSL certificate
     r'http://farsight-toolkit.ee.uh.edu/.*',
-    r'https://testng.org/*' # Invalid SSL certificate
+    r'https://testng.org/*', # Invalid SSL certificate
+    r'http://www.bio-rad.com/*' # 503 Server Error with Sphinx v1.8.5
 ]
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -857,7 +857,7 @@ reader = HitachiReader
 
 [I2I]
 extensions = .i2i
-developer = `Biomedical Imaging Group, UMass Medical School <http://invitro.umassmed.edu/>`_
+developer = `Biomedical Imaging Group, UMass Medical School <http://umassmed.edu/>`_
 bsd = no
 weHave = * several example datasets \n
 * a specification document \n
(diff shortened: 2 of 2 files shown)


sphinx/conf.py
```diff
@@ -432,5 +432,6 @@
     r'https://www.nis-elements.com/.*', # Invalid SSL certificate
     r'https://www.nikoninstruments.com/.*', # Invalid SSL certificate
     r'http://farsight-toolkit.ee.uh.edu/.*',
-    r'https://testng.org/*' # Invalid SSL certificate
+    r'https://testng.org/*', # Invalid SSL certificate
```
mtbc: These are regexps, not shell globs, hence the `.*` above. But do we even need a trailing "anything" or do these match part of the URL instead of necessarily the whole? Might be worth trying omitting that last bit altogether then re-triggering.

src/main/resources/format-pages.txt
```diff
@@ -857,7 +857,7 @@ reader = HitachiReader
 
 [I2I]
 extensions = .i2i
-developer = `Biomedical Imaging Group, UMass Medical School <http://invitro.umassmed.edu/>`_
+developer = `Biomedical Imaging Group, UMass Medical School <http://umassmed.edu/>`_
```
mtbc: The Biomedical Imaging Group is still at http://invitro.umassmed.edu/ so necessary to make the URL less specific?

src/main/resources/format-pages.txt
```diff
@@ -1002,9 +1002,9 @@ mif = true
 [Improvision Openlab Raw]
 extensions = .raw
 owner = `PerkinElmer <http://www.perkinelmer.com/>`_
-developer = `Improvision <http://www.perkinelmer.com/lab-products-and-services/cellular-imaging/index.html>`_
+developer = `Improvision <https://www.perkinelmer.com/lab-products-and-services/resources/software-downloads.html>`_
```
mtbc: Are these various Improvision things even still offered at all via PE's site?

dgault: There was mention of it a couple of days ago under the `cellular-imaging-software-downloads` link at the above address, however that link is currently dead. The site seems to be in a state of flux at the moment and is changing every few days, which is not ideal.

src/main/resources/format-pages.txt
dgault: Can you access it? That link had been dead and is still timing out for me.

sphinx/conf.py
dgault: I suspect you are right, probably worth a test at some point to see

src/main/resources/format-pages.txt
mtbc: Hmm, could just be worth waiting a bit, I'll drop them a line to ask when they'll be done.

src/main/resources/format-pages.txt
mtbc: I can. It's slow but it comes up. I'll drop them a line to warn them.

PR #136, opened 2019-11-19
For format information make list of wants conditional.

mtbc:
Pages like https://docs.openmicroscopy.org/bio-formats/6.3.0/formats/tiff.html with,

> We would like to have:

but then nothing after it look a bit strange so this PR omits that introductory clause.

### DIFF
--- a/src/main/resources/templates/FormatPage.vm
+++ b/src/main/resources/templates/FormatPage.vm
@@ -68,7 +68,9 @@ We currently have:
 $item
 #end
 
+#if ($weWant)
 We would like to have:
+#end
 
 #foreach ($item in $weWant)
 $item


mtbc:
Compare:
- https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-build-docs/jdk=JDK8,label=testintegration/ws/bio-formats-build/bio-formats-documentation/target/sphinx/html/formats/tiff.html
- https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-build-docs/jdk=JDK8,label=testintegration/ws/bio-formats-build/bio-formats-documentation/target/sphinx/html/formats/tiff.html

PR #137, opened 2019-12-11
Add mayo.edu to ignore list

dgault:
Adding to ignore list as it regularly fails link check with the below error:
broken    https://www.mayo.edu/research/core-resources/biomedical-imaging-resource-core/overview - 403 Client Error: ModSecurity Action for url: https://www.mayo.edu/research/core-resources/biomedical-imaging-resource-core/overview


dgault:
Link has not caused issues since.

PR #146, opened 2020-02-25
Adjust KLA Tencor surface profilers URL.

mtbc:
Hoping to improve [BIOFORMATS-build-docs](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-build-docs/).


mtbc:
Doesn't actually help with the broken link error but given their redirection probably worth leaving in anyway.

dgault:
No harm in having this included and given its redirecting makes sense

PR #148, opened 2020-03-10
Update path to bio-formats tag on image.sc

ctrueden:
Discourse changed from `/tags/...` to `/tag/...` recently. The old path still works, but you will only see the Bio-Formats sidebar description if you use the new path, so best to update it.


joshmoore:
Ah, thanks, @ctrueden. :+1: I saw that you said they were adding the other value, but didn't realize that the priority had swapped.

dgault:
New links all working correctly and the side bar is now showing

PR #151, opened 2020-03-19
Linkcheck fix

sbesson:
This aims at fixing long-running issues with the Bio-Formats documentation job.

For testing, the previous BIOFORMATS-build-docs CI job has been updated by a more straightforward BIOFORMATS-linkcheck one which consumes the Docker image built from https://github.com/ome/bio-formats-build and runs the linkcheck step. This should also use a Python 3 environment and thus benefit from the latest version of Sphinx

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
+---
 language: java
-
-sudo: false
-
-# http://docs.travis-ci.com/user/caching/#Arbitrary-directories
 cache:
   directories:
   - $HOME/.m2
+
+addons:
+  apt:
+    packages:
+      - python3-pip
 
 jdk:
   - openjdk11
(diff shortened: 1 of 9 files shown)


sbesson:
As discussed earlier, 5e22d6449e0d6ba03de9d458ea9dbf6a0f18478c should exclude all `https://github.com/ome/*` URLs from the linkcheck to reduce the spurious failures that have been affecting the docs builds since last week. /cc @mtbc

dgault:
Changes look good, and with the new environment and PR we actually have a green linkscheck build today https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/17/. Merging this now to bring the builds back to a state of stability

PR #153, opened 2020-04-06
Convert format-documentation page glossary into definition list

sbesson:
Sphinx 3.0.0 released yesterday throws warnings on duplicate glossary keys including lowercase variations (`bsd` vs `BSD`) - see https://github.com/sphinx-doc/sphinx/issues/7418

This has broken the daily [BIOFORMATS-image](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-image/360/) job


dgault:
Makes sense to remove glossary from `format-documentation` rather than `supported-formats` in that case

PR #157, opened 2020-04-14
Add MetaXpress format page

melissalinkert:
See discussion in https://github.com/openmicroscopy/data_repo_config/issues/452

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -1487,6 +1487,21 @@ notes = Commercial applications that support STK include: \n
 .. seealso:: \n
   `Metamorph imaging system overview <https://www.moleculardevices.com/systems/metamorph-research-imaging/metamorph-microscopy-automation-and-image-analysis-software>`_
 
+[MetaXpress]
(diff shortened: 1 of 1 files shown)


src/main/resources/format-pages.txt
```diff
@@ -1487,6 +1487,21 @@ notes = Commercial applications that support STK include: \n
 .. seealso:: \n
   `Metamorph imaging system overview <https://www.moleculardevices.com/systems/metamorph-research-imaging/metamorph-microscopy-automation-and-image-analysis-software>`_
 
+[MetaXpress]
+extensions = .htd, .tif, .tiff
+owner = `Molecular Devices <https://www.moleculardevices.com/>`_
+bsd = no
+weHave = * several MetaXpress datasets\n
+* `public sample datasets <https://downloads.openmicroscopy.org/images/CellWorX/>`__
```
sbesson, line 1495: The only question here is whether we are fine to point at a `CellWorX` URL until we sort out our samples internally. An alternative would be to create https://downloads.openmicroscopy.org/images/MetaXpress and redirect/proxy for now.

PR #164, opened 2020-06-09
Fix various hyperlinks.

mtbc:
Diffraction acquired SBIG and rearranged the site. SLF4J don't like HTTPS.


mtbc:
The linked zip contains files like `file_format.pdf`.

dgault:
Updated links look good, happy to move https://www.slf4j.org/ to the ignore list for now

PR #167, opened 2020-06-16
OME-TIFF: refer to commercial partners page

melissalinkert:
...instead of duplicating the list of commercial entities that support OME-TIFF.

The commercial partners page already has a ```#partnerships``` anchor to the ```Organizations supporting OME-TIFF``` section, so that is used here instead of anything new.


dgault:
Updated link is working as expected

PR #168, opened 2020-06-30
Temporarily add imagej.net to link check ignore list

dgault:
Adding imagej.net to the linkcheck ignore list until the site is fully restored. See https://forum.image.sc/t/imagej-wiki-is-down/39672 for further details


mtbc:
https://forum.image.sc/t/imagej-web-site-migration-timeline/40500 has some information, effort is ongoing.

PR #170, opened 2020-07-07
Fix misleading setFlattenedResolutions/setUnflattenedResolutions references

melissalinkert:
```setFlattenedResolutions``` works opposite to how it was documented, and ```setUnflattenedResolutions``` is not an API method.


dgault:
Updated document looks to be correct, lets get in included with 6.5.1

PR #165, opened 2020-06-11
Ignore LOCI hyperlinks.

mtbc:
This PR reduces [BIOFORMATS-linkcheck](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/) clutter while LOCI sort out their migration to https://eliceirilab.org/. *Do not merge,* they'll get back to us when they're sorted.

PR #178, opened 2020-08-14
Use canonical URL for Cytiva

sbesson:
The previous deep URLs fail the linkchecking using the latest Sphinx as HEAD requests lead to infinite redirects.

See 

```
$ curl -IL https://www.cytivalifesciences.com/en/us/shop/protein-analysis/molecular-imaging-for-proteins/imaging-systems
...
curl: (47) Maximum (50) redirects followed
```

vs 

```
$ curl -IL https://www.cytivalifesciences.com/
...
HTTP/1.1 200 OK
...
```

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -438,4 +438,6 @@ linkcheck_ignore = ['https://imspector.mpibpc.mpg.de',
     r'https://libjpeg-turbo.org',
     r'https://www.biovis.com/.*', # SSLV3_ALERT_HANDSHAKE_FAILURE
     r'https://github.com/ome/.*', # 429 too many requests for url
+    # Too many redirects - see https://github.com/sphinx-doc/sphinx/pull/8131
+    r'https://www.cytivalifesciences.com/en/us/.*',
 ]
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -94,7 +94,7 @@ mif = true
 
 [Amersham Biosciences Gel]
 extensions = .gel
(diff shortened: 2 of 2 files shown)


mtbc:
Interesting tradeoff here between providing a more useful URL for readers and trying to work around deficiencies in our CI. :grimacing:

sbesson:
I appreciate your point. To be fair and distribute the deficiencies, the usage of HEAD requests is fairly standard especially when it comes to link checking tools and these URLs unfortunately are unable to properly respond to such requests without infinite redirects (as shown by `curl -IL` above).

In terms of the change, I agree the old URLs are more informative than the canonical one. Happy to keep them but add them to the ignore list with a note (to be reverted) if that's the larger consensus.

I will separately look into the Sphinx logic for checking URLs and why the GET fallback does not work.

sbesson:
See also https://github.com/sphinx-doc/sphinx/pull/8131 which attempts at improving the linkcheck logic

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-push#551](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/551/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/551/consoleText) for more details.
Possible conflicts:
  - PR #165 mtbc 'Ignore LOCI hyperlinks.'
    - sphinx/conf.py
  - PR #168 dgault 'Temporarily add imagej.net to link check ignore list'
    - sphinx/conf.py

--conflicts

mtbc (approved):

PR #179, opened 2020-09-15
Various adjustments to follow LOCI site migration.

mtbc:
Links should be sensible and continue to work.


dgault:
The updated links and text all look to be correct and working. Merging for inclusion in 6.6.0

PR #180, opened 2020-12-09
Add page for Tecan Spark Cyto workspaces

melissalinkert:
See https://github.com/ome/bioformats/pull/3637.

Corresponding approved text for release notes is: ```Added a new reader for Tecan Spark Cyto workspace files, provided through a collaboration between Tecan Trading and Glencoe Software```


sbesson:
NB: I think in this case the Travis CI failures were legit as the new format page requires a `formats-gpl` JAr with the new reader included. I expect things to get green as the version of Bio-Formats gets bumped to 6.6.0

PR #181, opened 2020-12-10
Actions

jburel:
Replace travis by GHA
I have tested the deployment by pushing a tag to my fork
see job: https://github.com/jburel/bio-formats-documentation/actions/runs/413141191 

Screenshot below shows the doc on artifactory (I have deleted it since)
Screenshot 2020-12-10 at 14 23 26

Secrets will need to be set on this repository in order to deploy the doc via GHA
cc @sbesson


sbesson:
Great, a few comments:
- can you remove `.appveyor.yml` which is effectively covered by GitHub actions now?
- for the deployment, I have been considering the doc artifacts should be deployed as GitHub releases  rather than artifacts (and mid-term, this could even be published via readthedocs). Should we look into the switch? 
- could the release step running `mvn deploy` could be used for https://github.com/ome/bioformats to remove the remaining `.travis.yml` bit? 

/cc @dgault

jburel:
@sbesson https://github.com/ome/bio-formats-documentation/pull/181/commits/750ab59cb5cf8d0ce2af944dc1a0e163fa0948c4 removed the ``appveyor`` file

jburel:
@sbesson I am working on https://github.com/ome/bioformats now

jburel:
For this round I am more in favour of keeping things are they are since we still have some repo to look into
Then we can change, also using maven for the doc is not what our other repositories do.

PR #186, opened 2021-01-18
Temporary ignore https://spider.wadsworth.org

sbesson:
The SSL certificate expired last week. The SPIDER team has been contacted about
the update. In the meantime this should reduce the noise in the daily linkcheck
builds


dgault:
Merging as still an issue and no response from the initial contact

PR #187, opened 2021-02-09
exclude broken link

jburel:
exclude broken link wiki-bsse.ethz.ch


jburel:
https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/366/ is green
Merging so we can also have latest-ci green

PR #189, opened 2021-03-12
Add references to the public IMS formats specification page

sbesson:
Also link to the Imaris writer as the reference implementation for writing
IMS files

Following a support thread with EU Support BitPlane (Call#120669)


dgault:
New links look good and are correct. Merging for next patch release

PR #191, opened 2021-03-22
Rtd

jburel:
This PR makes the documentation available on readthedocs.
When in place, changes do the documentation can be reviewed on readthedocs using the ``Build pull requests for this project`` option.
If a change to the Java code is made,  a doc commit will happen via GHA
The build process has not been changed.
doc available at https://jburel-bf.readthedocs.io/en/latest/


joshmoore:
Few thoughts/questions:
 * A heads up that on pull you will get a `v6.6.1` tag. For future test tags, maybe use a suffix of some form?
 * What are the downsides of *not* including all the source code?
 * How do changes from ome/bioformats propagate here?

jburel:
I removed the tag. Thanks for pointing that out.

jburel:
* If follows the same logic has before, if a change is made, in the GHA, a build is run and the changes are committed to the master branch.
* Which source code are you referring to?

joshmoore:
> Which source code are you referring to?

The many lines that are added in this PR.

> If follows the same logic has before, if a change is made, in the GHA, a build is run and the changes are committed to the master branch.

Sorry, can you spell out which action/branch on which repo?

joshmoore:
> This could be changed to work with the snapshot if we want to 

Understood. I was trying to find how it was linked in this PR. Thanks.

> The mvn build generates .rst...

Did you mean to commit them?

sbesson:
Very initial reaction while looking at this PR is that I was also slightly put off by the number of committed files and lines of code. It certainly feels hard to review.

From a technology perspective, Sphinx uses Jinja2 as a templating engine and it feels like we it might be possible to leverage this and reduce the amount of redundant/unnecessary RST files. A good example of this is the `autodoc`  extension which effectively acts as a source generation step from the Python module and is compatible with readthedocs (https://omero-rois.readthedocs.io/en/latest/). I have not found whether a similar mechanism would exist that could apply to this use case, effectively generating pages from an INI/JSON/YML data file.

jburel:
I have used ``autodoc`` on other projects e.g. https://github.com/jburel/scripts/tree/docs_sphinx
I have not yet found a way of injecting the files for this specific use case without changing the build system. The usage of ``maven`` to generate the doc is certainly very specific to BF (only project in our stack using it!). So we might have to review that strategy.
I agree that the initial commits is sligtly off-putting.

jburel:
TODO: 
 * Examples Java/matlab/ini: use link to files on GH, No idea why they were copied in the first place from https://github.com/ome/bio-formats-examples/tree/master/src/main/java. Convenience?
 * Check at moving the target

sbesson:
> Examples Java/matlab/ini: use link to files on GH, No idea why they were copied in the first place from https://github.com/ome/bio-formats-examples/tree/master/src/main/java. Convenience?

That was another thought I had while reading the diff. I think the sources embedding largely comes from the historical monolithic repository. I personally would be happy for these links to point to the tagged sources in https://github.com/ome/bio-formats-examples. Happy to settle on this at the next weekly @ome/formats meeting

PR #196, opened 2021-03-24
RTD v2

jburel:
This PR replaces #191 
This PR makes the documentation available on readthedocs.
When in place, changes do the documentation can be reviewed on readthedocs using the ``Build pull requests for this project`` option.

doc available at https://jburel-bf.readthedocs.io/en/latest/

The ``compile`` phase of the build process using maven has been adjusted. This should not affect the local build using either the ``Makefile`` or ``maven`` build (tested via GHA). The set of changes allows the migration without changing the maven workflow that people use for building the doc locally or in GHA.


sphinx/developers/conversion.rst
```diff
@@ -7,8 +7,8 @@ using Bio-Formats version 4.2 and later.
 Working example code is provided in
 :download:`FileConvert.java <examples/FileConvert.java>` - code from that
 class is referenced here in part. You will need to have
-**bioformats_package.jar** in your Java CLASSPATH in order to compile
-FileConvert.java.
+:file:`bioformats_package.jar` in your Java :envvar:`CLASSPATH` in order to compile
```
sbesson, line 10: To keep this PR minimal and focus on the infrastructure changes required to publish on readthedocs, could we extract these changes to a separate PR?

jburel: yes. I made more changes initially required for RTD but I had to roll them back.

jburel: done

sphinx/conf.py
```diff
@@ -23,11 +21,34 @@
...
+    with open("../target/sphinx/conf.py") as f:
+        content = f.readlines()
+        for x in content:
+            if x.startswith(value):
+                values = x.split("=")
+                return values[1].strip().replace('\'', '')
+
+
+if os.environ.get('READTHEDOCS') == 'True':
+    os.system('mvn install -DskipSphinxTests=true -f ../pom.xml')
```
sbesson, line 39: Is there  a potential here for circular dependency since `mvn install` also wraps `sphinx-build`. Potentially only the relevant Maven phases should be invoked i.e. `generate-sources` ?

sphinx/conf.py
```diff
@@ -41,15 +62,10 @@
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.)
-extensions = ['sphinx.ext.extlinks', 'edit_on_github']
```
sbesson, line 44: This extensions is no longer necessary as it is built into readthedocs?

sbesson (commented):
Compared to the first attempt in #191, these changes are much more contained and feel less invasive. In particular there is no change to the way generated rst pages are built.

`conf.py` includes envvar conditional changes allowing to bootstrap the source generation before executing `sphinx-build`. As indicated via inline comments, my main concern is the potential for an infinite loop under some circumstances. Maybe we need to define which entry points are supported. I assume `mvn install` and`make html`? 

In terms of the generated documentation, https://jburel-bf.readthedocs.io/en/latest/ looks good to me and I think most of the content is there but I haven't done a detailed read yet. A few high-level questions:
- are we using readthedocs default theme? 
- what should be the publication URL?
- is every tag published or only latest URLs?

Waiting for feedback from the rest of @ome/formats. In terms of timeline, we just released Bio-Formats 6.6.1 and will be moving towards Bio-Formats 6.7.0 so now might actually be the best time to start publishing under readthedocs and invest into getting the documentation in a suitable state for all of us.


Otherwise, I could see some improvements especially in terms of expanding the table of contents in the left-hand menu like https://docs.readthedocs.io/. But this is largely outside the scope of this PR.

sphinx/conf.py
jburel: That should have been removed

sphinx/conf.py
jburel: I went for the "easiest" option at this stage of the work to see if it works, this needs to be broken up

jburel:
```
 * are we using readthedocs default theme?
 * what should be the publication URL?
* is every tag published or only latest URLs?
```
* We have used the default theme in other repositories. I have modified the summary table so it works with this theme. Changes theme will probably required a review
* URL: default domain using readthedocs.io or we customized https://docs.readthedocs.io/en/stable/custom_domains.html. Since we are migrating all our docs, it might be easier to use the default domain
* As mentioned previously we can choose to hide latest (merge PR) and only publish stable (tag) https://docs.readthedocs.io/en/stable/versions.html

sphinx/conf.py
jburel: clarification: ``generate-sources`` will not be enough

sphinx/conf.py
jburel: actually now this is required for ``mvn`` build if one is building locally

sphinx/conf.py
jburel: I will look into breaking it up. We also have ``zip`` step for example that we do not need

sphinx/_ext/.gitignore
```diff
@@ -1 +0,0 @@
-edit_on_github.pyc
```
joshmoore, line 1: Just a heads up that on a non-clean build, removal of this file can lead to `scc merge` failing since untracked files are present:

```
(base) /opt/own-bf/bio-formats-build/bio-formats-documentation $git status
HEAD detached from 8847d26f
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	sphinx/_ext/
```

sbesson: Maybe a case for adding `_ext` or `*.pyc` to the top-level `.gitignore`

sbesson:
I think the latest outage is a reminder of the proposal to migration of this documentation towards hosted resources.

One of my initial pondering was whether all published tags would be retained but this seems to be the case while looking at https://readthedocs.org/projects/sagemaker/versions/.

This leaves a few outstanding questions from my side mostly related to policy and process. I do not think all details need not be fully resolved in order to merge this PR but we should at least agree  on the progression

- do we have a consensu on publishing/exposing the `stable` documentation (i.e. last tag) vs `latest` (i.e. HEAD)?
  * `latest` allows to publish quickly e.g. a typo fix, a broken link but means that documentation corresponding to upcoming/unreleased features cannot be merged in the main development branch.
  * `stable` gives more control as a tag needs to be pushed to make a feature public. However, if we consider updating the stable documentation in-between Bio-Formats releases, the versions of the components will start diverging and we will need to define policies. I suspect we could enforce that the `major.minor` versions should always be consistent i.e. the version `x.y.0` tag of the documentation will match the `x.y.0` tag of Bio-Formats? 
-  Redirecting https://downloads.openmicroscopy.org/bio-formats/latest/ towards `stable/latest` should be easy. How should we handle the major/minor version versions redirects e.g. https://downloads.openmicroscopy.org/bio-formats/6?
- Would the previously published versions of the Bio-Formats documentation would remain on https://docs.openmicroscopy.org/bio-formats/ ?
- although it is not within this repository, what is the overarching plan for the Javadoc  API documentation? The latter is currently hosted under downloads.openmicroscopy.org but equally as critical as shown in https://forum.image.sc/t/openmicroscopy-org-down/60062/5. Dropping a few options:
  * use javadoc hosted solutions such as https://javadoc.io/ (per-component and require Maven Central)
  * try to integrate Javadoc into Sphinx a la `autodoc`

jburel:
* The release doc should point to ``stable`` as discussed previously. Since https://downloads.openmicroscopy.org/bio-formats/6 should redirect to the stable i.e. the doc pointing to the release package. The only downside of keeping doc and code in synch means that fixing a typo for example will not be possible on "stable". 
* For previous doc, we could still host under docs. That means that the release workflow should still have a step to push to docs.o.org. Another option could be to offer a zip attached to the tag on github
* As of the Javadoc, the point was raised during the doc review done a few months back. We do have some of our doc already on https://javadoc.io/ not all.  I think will be the most suitable path for our java packages but this is obviously a body of work. Considering no action was taken when raised the first time, it appears to be a lower priority. Outside the scope of this repository, this will not work for things like slice document (ICE)

sbesson:
> Since https://downloads.openmicroscopy.org/bio-formats/6 should redirect to the stable 

https://downloads.openmicroscopy.org/bio-formats/latest should redirect to the readthedoc stable but https://downloads.openmicroscopy.org/bio-formats/6 should redirect to the latest 6.x.y tag (e.g. when we release Bio-Formats 7) and e.g. https://downloads.openmicroscopy.org/bio-formats/6.7 should redirect to the latest 6.7.x tag. As far as I can tell there is no such logic on RTD

> For previous doc, we could still host under docs. That means that the release workflow should still have a step to push to docs.o.org.

No objection to keeping the old docs where they are, I think that's what we did previously. For new doc releases, what's  the benefit of keeping the dual workflow as I thought one of the advantage was to drop our own Jenkins-based infrastructure as well?

>  I think will be the most suitable path for our java packages but this is obviously a body of work.

Overall, I like javadoc.io but highlighting a few downsides:
1. only works for Java components as you mention
2. artifacts have to be hosted on Maven Central which requires at minimum a review of the dependencies (which should also be on Maven Central), a `groupId` change to `org.openmicroscopy` and a review of the release process
3. javadoc.io only provides component-level Javadoc while the [current Bio-Formats Javadoc](https://downloads.openmicroscopy.org/bio-formats/6.7.0/api/) combines `formats-api`, `formats-bsd` and `formats-gpl` (and the hyperlinks to e.g. `ome-xml` are already missing).

jburel:
> > Since https://downloads.openmicroscopy.org/bio-formats/6 should redirect to the stable
I misunderstood your comment. One option will be to change the redirect to the the "old" docs when a new release happens.

Pushing to docs.o.org to keep the old docs will have to happen at some stage, either automatically or manually.

sbesson:
Maybe we need to work through a concrete example. Assuming we agree to switch for 6.8.0, with the Bio-Formats binaries released as previously and a `v6.8.0` tag is pushed to this repository:

1. I assume the tag will get automatically built by RTD and the documentation would be published as https://bio-formats.readthedocs.io/en/v6.8.0/ and https://bio-formats.readthedocs.io/en/stable/ (until a new tag is pushed)
2. https://docs.openmicroscopy.org/bio-formats/latest/ could be adjusted to redirect to https://bio-formats.readthedocs.io/en/stable/ 
3. would you also propose to mirror the documentation and host  https://docs.openmicroscopy.org/bio-formats/6.8.0? or have this URL redirected to  https://bio-formats.readthedocs.io/en/v6.8.0/ ?
4. should https://docs.openmicroscopy.org/bio-formats/6/ and https://docs.openmicroscopy.org/bio-formats/6.8 be maintained?

jburel:
@sbesson I was incorrect. We can host older version of the docs on rtd. I specified a tag v5.6.2 https://jburel-ome-documentation.readthedocs.io/en/v5.6.2/ layout is not great but that's not the point. 
Moving forward we could possibly host "older version" there

jburel:
> 1. I assume the tag will get automatically built by RTD and the documentation would be published as https://bio-formats.readthedocs.io/en/v6.8.0/ and https://bio-formats.readthedocs.io/en/stable/ (until a new tag is pushed)

Stable will point to the latest tag and will be built automatically and available at https://bio-formats.readthedocs.io/en/stable/ but https://bio-formats.readthedocs.io/en/v6.8.0/ will have to be set to ``Active`` for the doc to be built. cf rtd doc

> 2. https://docs.openmicroscopy.org/bio-formats/latest/ could be adjusted to redirect to https://bio-formats.readthedocs.io/en/stable/
Yes

> 3. would you also propose to mirror the documentation and host  https://docs.openmicroscopy.org/bio-formats/6.8.0? or have this URL redirected to  https://bio-formats.readthedocs.io/en/v6.8.0/ ?
After setting the flag to ``Active`` cf. 1. This should redirect https://bio-formats.readthedocs.io/en/v6.8.0/ 
> 4. should https://docs.openmicroscopy.org/bio-formats/6/ and https://docs.openmicroscopy.org/bio-formats/6.8 be maintained?
If there are maintained, they should be updated every time there is a release in the 6 and 6.8 series.
I am not convinced we want to keep them

sbesson (approved):
From my side, this contains the minimal infrastructure for migrating the hosting of the technical documentation onto readthedocs. GIven the ongoing conversation, I expect there is enough infrastructure to support our versioned workflow and I strongly vote for getting rid of all the unnecessary steps so that a documentation release simply happens by bumping the version and pushing a tag on this repository.

As mentioned in the PR conversation, there will be a series of actions including setting up the documentation, testing configuration changes to redirect the various https://docs.openmicroscopy.org/bio-formats/ URLs to the new hosted site and potentially reviewing our usage of these URLs in various places such as the website. 

If the goal is to get this released together with Bio-Formats 6.8.0, I strongly vote for testing a documentation release workflow e.g. via pre-releases. @jburel @dgault we should probably discuss at the Formats meeting who will be in charge and how to coordinate this effort with the rest of the Bio-Format 6.8.0 roadmap

sphinx/_ext/.gitignore
jburel: done in ddd3721e937c1c270729dec0b9d987cbf34b578d

jburel:
Merging and preparing the doc

jburel:
The approach taken 9a5eae9cffcb1645d990ad895f4b511c4c619f47 does not seem to work anymore, I have reverted it for now so we can have the docs in place cf. https://bio-formats.readthedocs.io/en/latest/
Looking into the issue

jburel:
dea2c4a73d5ab2b0ca9cdcf02ee132aebc193775 fixes the table issue

sbesson:
The formats table still does not display the ratings for me. https://bio-formats.readthedocs.io/en/latest/supported-formats.html

jburel:
Screenshot 2023-01-25 at 09 20 55

sbesson:
:+1: I get the same result from an incognito window. I might have experienced a caching issue.

PR #197, opened 2021-03-25
fix format

jburel:
Adjust formatting. Initially in #96 
cc @sbesson


sbesson (approved):

jburel:
@dgault For example, this is a fix that will be immediately publicly available after the migration to rtd without a release.

PR #198, opened 2021-05-07
Update link to 2005 publication about the OME data model

sbesson:
The former link seems to be unavailable and it is unclear whether is will
remain supported. The DOI should resolve unconditionally moving forward


dgault:
New link looks good and passes linkchecks

PR #203, opened 2021-07-14
Update license of CellH5, BDV and KLB readers

sbesson:
These readers will be relicensed under BSD-2 as part of Bio-Formats 6.7.0


dgault:
Updates look good, merging for 6.7.0

PR #204, opened 2021-07-26
Fix link to Micro-Manager user guide anchor

sbesson:
Small update following the migration of the Micro-Manager website https://forum.image.sc/t/new-micro-manager-website-launched/55651


dgault:
Old link goes to the main Micro-Manager User Guide page, the updated link correctly goes to the Files on Disk anchor.

PR #206, opened 2021-08-12
Remove ImageJ mailing list archive link

sbesson:
imagej.1557.x6.nabble.com has been down for the last few days. Since this archive link is available from the imagej.nih.gov landing page and we do not have any other deep link to the archive, this commit proposes to remove the archive URL


sbesson:
See also https://list.nih.gov/cgi-bin/wa.exe?A2=IMAGEJ;610524d.2108

PR #213, opened 2021-09-20
Remove broken link to xuvtools.org

dgault:
This PR is in response to broken builds (https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/601/) which have highlighted that the http://www.xuvtools.org link appears to be no longer maintained.

@emmenlau would you perhaps know if there is a suitable alternative link we could use here?

### DIFF
--- a/sphinx/users/xuvtools/index.rst
+++ b/sphinx/users/xuvtools/index.rst
@@ -1,6 +1,6 @@
 XuvTools
 ========
 
-`XuvTools <http://www.xuvtools.org>`_ is automated 3D stitching software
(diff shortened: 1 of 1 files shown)


emmenlau:
Hi @dgault , I've fixed the webserver so that there should be no more errors accessing the website. However I've seen that there is another error in the database so that the website content is currently hidden. I will fix this in the close future, is that ok?

dgault:
Thats perfect, I will close this PR as the errors are no longer being seen

PR #231, opened 2022-02-15
Add version history for v6.9.0

dgault:
Adding the initial proposed version history section for the 6.9.0 release. The text from this PR will be the same text used on the release announcement on the OME website and community sites/mailing lists such as image.sc

The full list of PR's included will be from https://github.com/ome/bio-formats-documentation/milestone/20 and https://github.com/ome/bioformats/milestone/83?closed=1

### DIFF
--- a/sphinx/about/whats-new.rst
+++ b/sphinx/about/whats-new.rst
@@ -16,9 +16,6 @@ New file formats:
 
 File format fixes and improvements:
 
-* Aperio SVS TIFF
-   - fixed handling of macro and label images
-
 * FEI TIFF
    - fixed parsing of physical pixel sizes for Phenom data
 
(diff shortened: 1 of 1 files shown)


sphinx/about/whats-new.rst
```diff
@@ -1,6 +1,58 @@
...
+     This functionality was implemented with thanks to Leica Microsystems
+
+* Leica XLEF
+   - added support for reading Leica XLEF files.
+     This functionality was implemented with thanks to Leica Microsystems
+
+File format fixes and improvements:
+
+* Aperio SVS TIFF
+   - fixed handling of macro and label images
```
sbesson, line 20: I thought this was https://github.com/ome/bioformats/pull/3764 but it looks like this got excluded and pushed away?

sbesson (changes requested):
One outstanding question re the SVS change

sphinx/about/whats-new.rst
dgault: Yeah, apparently I only removed it from the website PR for some reason. Corrected now

sbesson (approved):

PR #237, opened 2022-03-15
Remove remaining references to QA uploads

melissalinkert:
Noticed while looking at https://github.com/glencoesoftware/bioformats2raw/issues/139.


dgault:
Changes look good and links are all working correctly.

PR #244, opened 2022-05-15
Update link to View5D

dgault:
Update link to View5D as previous link remains broken.
The suitable options I could find were between https://github.com/fiji/View5D/ and https://github.com/RainerHeintzmann/View5D. As the content relates to users using the plugin in ImageJ I opted for the FIJI link.

### DIFF
--- a/sphinx/users/imagej/index.rst
+++ b/sphinx/users/imagej/index.rst
@@ -27,7 +27,7 @@ ways:
    `Image5D <http://imagej.net/Image5D>`_ plugin
    (if installed)
 -  With Rainer Heintzmann's
-   `View5D <http://www.nanoimaging.de/View5D/>`_ plugin (if installed)
+   `View5D <https://github.com/fiji/View5D/>`_ plugin (if installed)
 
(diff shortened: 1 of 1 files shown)


sphinx/users/imagej/index.rst
```diff
@@ -27,7 +27,7 @@ ways:
    `Image5D <http://imagej.net/Image5D>`_ plugin
    (if installed)
 -  With Rainer Heintzmann's
-   `View5D <http://www.nanoimaging.de/View5D/>`_ plugin (if installed)
+   `View5D <https://github.com/fiji/View5D/>`_ plugin (if installed)
```
sbesson, line 30: Another possible option might be https://imagej.net/plugins/view5d which has the benefit to linking to both the source code repository and the (currently broken) homepage

sbesson (approved):
One other alternative link mentioned in https://github.com/ome/bio-formats-documentation/pull/244#discussion_r873212181 but I don't feel strongly either way. The proposed suggestion is equally valid and fixes the broken link

dgault:
Agreed, that is probably a better option as it links to both the old homepage and the source code

PR #249, opened 2022-05-27
Add new format page for Slidebook 7

dgault:
Add a new format page for the new Slidebook 7 Reader implemented in https://github.com/ome/bioformats/pull/3731

Ratings are based on https://docs.openmicroscopy.org/bio-formats/6.9.1/supported-formats.html#term-ratings-legend-and-definitions


src/main/resources/format-pages.txt
```diff
@@ -30,6 +30,21 @@ http://www.openmicroscopy.org/info/slidebook for details.\n
 \n
 .. _Intelligent Imaging Innovations: http://www.intelligent-imaging.com/
 
+[3i Slidebook 7]
+extensions = .sldy
+owner = `Intelligent Imaging Innovations`_
+developer = `Intelligent Imaging Innovations`_
+bsd = yes
+software = `Slidebook <https://www.intelligent-imaging.com/slidebook/>`_
+weHave = * a small number of datasets \n
```
sbesson, line 39: Possibly unrelated to this PR, can we make our sample files public?

dgault: The sample files were submitted publicly on the PR so that should be no problem. Do you want that included with this PR or can that be handled as a follow up?

sbesson: A follow-up PR is absolutely fine. I just happened to notice this but it should neither block the new page nor the release.

sbesson (approved):

PR #250, opened 2022-05-30
Update extlinks to comply with Sphinx 5 deprecation warnings

sbesson:
The [release of Sphinx 5.0.0](https://github.com/sphinx-doc/sphinx/releases/tag/v5.0.0) has caused our CI infrastructure to fail due to the deprecation warnings for the extlinks.

Investigating quickly, it looks like the remediation is fairly straightforward and the display name should either be `None` or a string containing a single occurence of `%s`. This PR should demonstrate the relevant fix on the BIo-Formats documentation.

There is a limited number of places using extlinks across our code - see https://github.com/search?q=org%3Aome+extlinks&type=code. Unless other features have been deprecated/dropped in Sphinx 5, an option would be to review and adjust all these to use the new syntax and validate it fixes the build with Sphinx 5.
The alternative as we have done previously would be to cap "Sphinx<5"

@jburel similar considerations will apply to the OMERO reference documentation and guides

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -127,55 +127,55 @@ imagesc_root = 'https://forum.image.sc'
 
 extlinks = {
     # image.sc
-    'imagesc' : (imagesc_root + '/%s', '#'),
+    'imagesc' : (imagesc_root + '/%s', '#%s'),
     # Trac links
-    'ticket' : (trac_root + '/ticket/%s', '#'),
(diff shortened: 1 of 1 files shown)


dgault:
The changes look fine to me, if we can get builds green again with the new syntax then I would happy to get the changes in rather then capping the version.

sbesson:
Looks like the re-execution of the nightly CI builds with this PR included completed successfully.

dgault (approved):

PR #248, opened 2022-05-27
Add version history for v6.10.0

dgault:
Adding the initial proposed version history section for the 6.10.0 release.

The full list of PR's included will be from https://github.com/ome/bioformats/milestone/86?closed=1

PR #251, opened 2022-05-30
Rename Vectra QPTIFF format page

dgault:
Renaming the Vectra QPTIFF format page following on from the discussion on https://github.com/ome/bio-formats-documentation/pull/248
The public sample docs and public spec are already under Vectra QPTIFF so shouldn't need to be updated

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -2122,22 +2122,6 @@ Commercial applications that support this format include: \n
 .. seealso:: \n
   `PerkinElmer UltraVIEW system overview (pdf) <https://www.perkinelmer.com/lab-solutions/resources/PDFs/LST/Brochures/BRO_UltraVIEW-VoX-Product-Brochure.pdf>`_
 
-[PerkinElmer Vectra QPTIFF]
-extensions = .tif, .qptiff
-owner = `PerkinElmer <https://www.perkinelmer.com/>`_
-bsd = no
-pyramid = yes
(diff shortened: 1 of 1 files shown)


sbesson (commented):
With this PR, the documentation now generates the following page:

Screenshot 2022-05-30 at 15 35 54

From my side, this is a faithful representation of the state of this format, which captures both  the former and current owner, preserving the original instrument for consistency with the reader and curated repository names. 

Two immediate questions come to mind and might require @melissalinkert's knowledge:

-   should the note be amended to mention that support was added in partnership with PerkinElmer as well Akoya Biosciences?
- https://github.com/ome/bioformats/pull/3829 mentions a schema version 4, can we go as far as expressing a supported range via the `versions` key?

pwalczysko:
Builds well on https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-build/jdk=JDK8,label=testintegration/lastSuccessfulBuild/artifact/bio-formats-build/bio-formats-documentation/target/

Linkage works, makes sense. The only comment is on the "specification document" - if it looks as expected, then, after the change requested by @jrswedlow is implemented, ready to merge fmpov.

pwalczysko:
Last commit looks sensible, see screenshot
Screenshot 2022-05-31 at 12 43 01
Link works. Build works too https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-build/jdk=JDK8,label=testintegration/lastSuccessfulBuild/artifact/bio-formats-build/bio-formats-documentation/target/

sbesson (approved):

PR #253, opened 2022-05-31
Bump Bio-Formats version to 6.10.0

dgault:
Bumping BF version to 6.10.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
   <properties>
     <ome-common.version>6.0.6</ome-common.version>
     <ome-model.version>6.2.2</ome-model.version>
-    <bioformats.version>6.9.1</bioformats.version>
+    <bioformats.version>6.10.0</bioformats.version>
     <formats-gpl.version>${bioformats.version}</formats-gpl.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -25,7 +25,7 @@
   <properties>
     <ome-common.version>6.0.6</ome-common.version>
```
sbesson, line 26:
```suggestion
    <ome-common.version>6.0.9</ome-common.version>
```

pom.xml
```diff
@@ -25,7 +25,7 @@
...
     <ome-model.version>6.2.2</ome-model.version>
```
sbesson, line 27:
```suggestion
    <ome-model.version>6.3.1</ome-model.version>
```

sbesson (commented):
Two suggestions to update the version of the low-level components as well since they are being bumped. Otherwise, reads well to me

pom.xml
dgault: Sorry, missed this comment, follow PR open now: https://github.com/ome/bio-formats-documentation/pull/256

PR #258, opened 2022-06-23
Add visitech.co.uk to the exclude list (invalid SSL certificate)

sbesson:
The certificate expired lately and has not been fixed yet


sbesson:
Closing in favor of #259

PR #265, opened 2022-08-18
Zeiss AxioVision - Remove software links

dgault:
PR to fix broken links in linkcheck test: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/944/

The AxioVision software is now classed as End of Support (see https://www.zeiss.co.uk/microscopy/local-content/2021/zeiss-axiovision-software-end-of-support.html) and no longer appears on the list of software products at https://www.zeiss.com/microscopy/en/products/software.html

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -2752,10 +2752,9 @@ mif = true
 pagename = zeiss-axiovision-zvi
 extensions = .zvi
 owner = `ZEISS International <https://www.zeiss.com/corporate/int/home.html>`_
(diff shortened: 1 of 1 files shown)


sbesson (commented):
Thanks @dgault. The removal of the dead links looks good to me and reflects the state of support of the software.

 @melissalinkert might have other suggestions but a possible update would be to add a note for this format page linking to the EOL announcement.

melissalinkert:
Agreed that linking the EOL announcement makes sense. ZEN can open .zvi files, so might be worth saying that as well in the `software` section.

sbesson (approved):

PR #266, opened 2022-09-08
Add footnote on setting fake file channel colors

melissalinkert:
See https://github.com/ome/bioformats/blob/develop/components/formats-bsd/src/loci/formats/in/FakeReader.java#L1409


dgault:
Changes look good and read well, merging for next release

PR #268, opened 2022-10-03
Add link to public samples generated by Olympus FluoView systems

sbesson:
Includes the sample uploaded in the context of https://github.com/ome/bioformats/pull/3874


dgault (approved):
Link looks good

PR #279, opened 2023-01-24
fix injection of css

jburel:
Fix table issues


jburel:
It seems that the webhook did not work

jburel:
see https://jburel-bf.readthedocs.io/en/latest/supported-formats.html

PR #281, opened 2023-01-25
remove sphinx following migration to rtd

jburel:
Doc is built using Readthedocs


jburel:
Everything is in place but it does not seem to work. Closing and I will review that later

PR #284, opened 2023-02-08
remove sphinx following migration to rtd

jburel:
Doc should now be built by rtd

This will reduce the build time


dgault (approved):

dgault:
Removing the sphinx workflow should now be safe

PR #290, opened 2023-02-21
Add version support to LI-Flim format page

dgault:
As a follow up to https://github.com/ome/bioformats/pull/3936, adding version info to the format page as both version 1.0 and 2.0 are now supported

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -1313,6 +1313,7 @@ reader = KodakReader
 extensions = .fli
 developer = `Lambert Instruments <https://www.lambertinstruments.com>`_
 bsd = no
+versions = 1.0, 2.0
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Based on the output of https://bio-formats--290.org.readthedocs.build/en/290/formats/lambert-instruments-flim.html

jburel:
The doc could be pushed before the next release. 
If this is the case, tag as v6.12.0-1 see for example https://github.com/ome/omero-documentation/releases/tag/v5.6.6-3

sbesson:
> The doc could be pushed before the next release.
In that specific case, wasn't the corresponding feature only merged after the 6.12.0 release? In which case, this should definitely not be published before the next Bio-Formats release.

jburel:
It is the case, ignore my comment.
I thought it was something missed from the release

PR #291, opened 2023-03-06
Update link for ZEISS AxioVision end of support

dgault:
Updating the link due to link check failures: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/1140/consoleFull

Previously the link was to an announcement of end of support for AxioVision. The new link is to the list of discontinued products.


sbesson (approved):
This link seems to keep on changing but the new landing page for all discontinued products seems as good as any

PR #292, opened 2023-03-20
Link version history from main page

melissalinkert:
Include the version history page on the front page and in the main table of contents. Without this PR, `about/whats-new` is only linked from `about/index`, and isn't easily visible within that page. With this PR, the version history should be more visible, especially for anyone who doesn't already know where to find it.

In looking at this, I also noticed that the `toctree` in `about/index` results in some weird nesting in the table of contents. The version history ends up as a sub-section under `Bio-Formats metadata processing`, when I would have expected it to be its own section. This means the version history was even harder to find. I did not attempt to address this problem here as I believe it is the same as what is described in https://github.com/sphinx-doc/sphinx/issues/3007.

### DIFF
--- a/sphinx/about/index.rst
+++ b/sphinx/about/index.rst
@@ -194,10 +194,3 @@ metadata, original metadata, and OME metadata.
    bug free, but we are constantly working to improve it. We would
    greatly appreciate any and all input from users concerning missing or
    improperly converted metadata fields.
-
-.. toctree::
-    :maxdepth: 1
-    :hidden:
-
-    bug-reporting
-    whats-new
--- a/sphinx/about/whats-new.rst
+++ b/sphinx/about/whats-new.rst
@@ -1,7 +1,7 @@
 Version history
 ===============
 
-6.12.0 (2022 February)
+6.12.0 (2023 February)
 ----------------------
 
 File format fixes and improvements:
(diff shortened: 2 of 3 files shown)


jburel:
the change makes sense
Could you also fix the date in the ``whats-new`` page 
i.e.`` 6.12.0 (2022 February) `` => `` 6.12.0 (2023 February) ``

sbesson (approved):
Big 👍 , I had the same realisation with the version history being buried but never took the time to fix it properly.
Talking to several end-users/consumers, they emphasised how useful the information on this page was. So I am totally supportive of the decision chosen here which adds a `Version history` entry to the left-hand navigation.

sbesson:
Regarding the hidden toctree, note the two sections still appear as nested under `Bio-Formats metadata processing`. I understand if we want this separately but overall I wonder if 1- `Reporting a bug` should be treated similarly to `Version history` and moved to the top-level, 2- the hidden toc tree should be simply deleted.

melissalinkert:
> Regarding the hidden toctree, note the two sections still appear as nested under `Bio-Formats metadata processing`.

That was expected - I was mostly noting for completeness, the first 2 commits didn't try to fix that.

>  I understand if we want this separately but overall I wonder if 1- `Reporting a bug` should be treated similarly to `Version history` and moved to the top-level, 2- the hidden toc tree should be simply deleted.

bf6ceb0 and f63492e should do this.

sbesson (approved):
Thanks @melissalinkert, personally I am completely happy with the state of https://bio-formats--292.org.readthedocs.build/en/292/ but I'll leave others to review.

jburel:
it looks good to me too

dgault (approved):

PR #295, opened 2023-04-07
Add documentation for Bioconductor RBioFormats

dgault:
This is a follow up to the request raised on forum thread https://forum.image.sc/t/possible-addition-to-https-www-openmicroscopy-org-bio-formats-downloads-rbioformats/79474

This PR adds 2 new pages, adding Bioconductor to the Users list (under libraries and sciprting applications) and also a page on `Using Bio-Formats in R` under the developer section.

### DIFF
--- a/sphinx/developers/r-dev.rst
+++ b/sphinx/developers/r-dev.rst
@@ -24,3 +24,5 @@ to allow reading of proprietary microscopy image data and metadata.:
 
     # To view documentation for the version of this package installed in your system
     browseVignettes("RBioFormats")
+
+For further details and examples of how to use the package, please see the `API documentation<https://bioconductor.org/packages/devel/bioc/vignettes/RBioFormats/inst/doc/RBioFormats.html>`_ 
--- a/sphinx/developers/r-dev.rst
+++ b/sphinx/developers/r-dev.rst
@@ -1,0 +1,26 @@
+Using Bio-Formats in R
+======================
+
+OME does not currently provide an R implementation for Bio-Formats.
+However, there are options provided by the community:
+
+Bioconductor RBioFormats
+-------------------------
+
+The `RBioFormats <https://bioconductor.org/packages/devel/bioc/html/RBioFormats.html>`_ 
+project from Bioconductor provides an R package which interfaces the OME Bio-Formats Java library 
+to allow reading of proprietary microscopy image data and metadata.:
+
+.. code-block::
+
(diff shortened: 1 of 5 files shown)


sphinx/developers/r-dev.rst
```diff
@@ -0,0 +1,26 @@
...
+      install.packages("BiocManager")
+
+    # The following initializes usage of Bioc devel
+    BiocManager::install(version='devel')
+
+    BiocManager::install("RBioFormats")
+
+
+    # To view documentation for the version of this package installed in your system
+    browseVignettes("RBioFormats")
```
sbesson, line 26: Possibly the dev section could include a reference to https://bioconductor.org/packages/devel/bioc/vignettes/RBioFormats/inst/doc/RBioFormats.html which is the API documentation

sbesson (commented):
Thanks @dgault, the addition of two separate pages (user & dev) makes sense to me and follows the established pattern for e.g. MATLAB and other applications - see https://bio-formats--295.org.readthedocs.build/en/295/users/index.html and https://bio-formats--295.org.readthedocs.build/en/295/developers/index.html.

My primary question is whether the user page should be called `Bioconductor` under `Libraries and scripting applications` or `R` under `Numerical data processing applications` e.g. by symmetry with MATLAB or Octave.

dgault:
I can see either option making sense for the user page, happy to move it to `R` under `Numerical data processing applications` if that is preferred.

melissalinkert:
Maybe worth adding a statement to the Bioconductor user page similar to what's at the top of the dev page (so it's clearer that this isn't something OME develops). Otherwise looks fine to me, I don't have strong opinions on which section to use for the user page.

sphinx/developers/r-dev.rst
```diff
@@ -0,0 +1,28 @@
...
+
+For further details and examples of how to use the package, please see the `API documentation<https://bioconductor.org/packages/devel/bioc/vignettes/RBioFormats/inst/doc/RBioFormats.html>`_
```
sbesson, line 28:
```suggestion
For further details and examples of how to use the package, please see the `API documentation <https://bioconductor.org/packages/devel/bioc/vignettes/RBioFormats/inst/doc/RBioFormats.html>`_ 
```

Looks like you need an extra space to get this rendered as an hyperlink.

sbesson (changes requested):
One rendering issue. Otherwise looks like good to me

sbesson (approved):

PR #298, opened 2023-04-19
Add -fill option for showinf/bfconvert

melissalinkert:
See https://github.com/ome/bioformats/pull/3963#pullrequestreview-1391963438


dgault (approved):
Docs look good as per https://bio-formats--298.org.readthedocs.build/en/298/users/comlinetools/display.html and https://bio-formats--298.org.readthedocs.build/en/298/users/comlinetools/conversion.html

PR #306, opened 2023-06-16
Remove notes about sorting table columns

melissalinkert:
This feature has been broken for many years: https://trello.com/c/oNX1t6DS/245-formats-table-columns-not-sortable

A quick look through RTD documentation suggests there isn't an easy way to re-add this feature, so for now just removing the `You can sort...` statements for clarity.


dgault (approved):
Makes sense to just remove the references

PR #312, opened 2023-07-03
Set sphinx.fail_on_warnings to true

sbesson:
See https://github.com/ome/bio-formats-documentation/pull/311/commits/2643cd03e1f11acdbf986210216d2c9fb670989a

The initial commit in https://github.com/ome/bio-formats-documentation/pull/311 passed the readthedocs CI builds but failed the OME Jenkins CI jobs. This is due to a difference in the the handling of warnings with OME Jenkins CI being the strictest of the two.

To reduce the confusion, this sets the appropriate readthedocs configuration `sphinx.fail_on_warning` to `true`.

This PR is built on top of 698f941a6b4ad4343b92512f3dfe3eaf2ab62d44 so the expectation is that the first build will fail. With 2643cd03e1f11acdbf986210216d2c9fb670989a the second run should be successful.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,6 +8,7 @@ version: 2
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: sphinx/conf.py
+  fail_on_warning: true
 
 # Build documentation with MkDocs
 #mkdocs:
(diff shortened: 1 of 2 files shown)


sbesson:
See https://readthedocs.org/projects/bio-formats/builds/21187543/ vs https://readthedocs.org/projects/bio-formats/builds/21187633/ for the effect of the flag with and without an invalid Sphinx syntax.
Happy to close this PR in favor of #311 which now also contains the configuration change or alternatively separate the documentation and configuration changes as two separate PRs.

jburel:
Happy to have the commit included in #311

PR #318, opened 2023-07-13
Remove Woolz format

dgault:
Removes references of the Woolz format from the documentation

### DIFF
--- a/sphinx/about/whats-new.rst
+++ b/sphinx/about/whats-new.rst
@@ -3110,7 +3110,7 @@ Java bug fixes:
      Jean-Yves Tinevez)
    - :doc:`LaVision Imspector </formats/lavision-imspector>`
    - :doc:`PCORAW </formats/pcoraw>`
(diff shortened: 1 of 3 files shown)


jburel:
`` - :doc:`Woolz </formats/woolz>` (thanks to Bill Hill)`` will break

sbesson (commented):
From the generated documentation, the only two locations I could find which still reference Woolz are:
- the [dataset structure table](https://bio-formats--318.org.readthedocs.build/en/318/formats/dataset-table.html)
- the [WlzReader metadata page](https://bio-formats--318.org.readthedocs.build/en/318/metadata/WlzReader.html)

I think both pages are auto-generated from the sources and should automatically get updated/removed when the dependency of this documentation repository is updated to `formats-gpl:7.0.0` ?

melissalinkert (approved):

sbesson (approved):
Assuming we start capturing the checks for the BF 7 release elsewhere

PR #321, opened 2023-07-20
Remove references to LuraWave

dgault:
This is a companion PR to https://github.com/ome/bioformats/pull/4040

We may also want to update https://github.com/ome/ome-stubs to remove the LuraWave stubs


melissalinkert (approved):
Looks good to me.

PR #325, opened 2023-07-28
Add dicom.dual_personality writer option

dgault:
Adding documentation for the new option in https://github.com/ome/bioformats/pull/3911


sbesson:
Deployed at https://bio-formats--325.org.readthedocs.build/en/325/formats/options.html and looks good. Only question is whether we want to shorten the hyperlink. But deferring to you and @melissalinkert as I won't be able to perform additional review before the Bio-Formats 7.0.0 release

PR #330, opened 2023-08-18
update uri

jburel:
Point uri to ome_model_doc https://ome-model.readthedocs.io/en/stable/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,12 +36,9 @@
     <sphinx.ome_source.user>ome</sphinx.ome_source.user>
     <sphinx.openmicroscopy_source.user>openmicroscopy</sphinx.openmicroscopy_source.user>
 
-    <ome-model.uri>https://ome-model.readthedocs.io/en/stable/</ome-model.uri>
-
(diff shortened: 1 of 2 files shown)


sbesson:
Do you foresee the need to override this base URI ever in the near future?
Otherwise, an alternative would be to completely get rid of this variable and its usage in https://github.com/ome/bio-formats-documentation/blob/ab5e213dbf1b5ef9e864ecc2e3a99bc67d83e67c/sphinx/conf.py#L217-L218 and update https://github.com/ome/bio-formats-documentation/blob/ab5e213dbf1b5ef9e864ecc2e3a99bc67d83e67c/sphinx/conf.py#L170 to use the new base URI

jburel:
I think the  BF linkcheck job could use ``latest`` (merge  PR) instead of ``stable`` (release/tag). That will match what was previously in place

jburel:
Discussed during the weekly BF call, we decided not to keep the variable

dgault (approved):
Changes look good from my side

PR #369, opened 2024-04-10
Describe new `-quality` option in `bfconvert`

melissalinkert:
See https://github.com/ome/bioformats/pull/4171#pullrequestreview-1991041388


sbesson (approved):
Reads well to me.
I had missed the fact that `ome.codecs.JPEGCodec` was internally updating the quality within the [0.25,1.0] range so it's definitely useful to have that explicit in some user-facing documentation. As a follow-up, I also wonder if the codec should log a warning statement when this value is modified internally

PR #376, opened 2024-06-04
Remove link to barre.dev samples

dgault:
Temporarily removing the link to Sebastien Barre's DICOM samples as the entire site has been down for a number of days now: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/82/consoleFull

I haven't been able to find a suitable alternative link for these samples. I will monitor to see if the site returns, if so then we can close this PR


melissalinkert (approved):
I don't know of a replacement either, so makes sense to remove.

PR #378, opened 2024-06-10
Update ImageJ links to use imagej.net

dgault:
Updating the Image J links to use the correct imagej.net instead of the old image.nih.gov
Failures have been seen in the daily linkcheck: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/88/consoleFull

### DIFF
--- a/sphinx/users/imagej/index.rst
+++ b/sphinx/users/imagej/index.rst
@@ -1,7 +1,7 @@
 ImageJ overview
 ===============
 
-`ImageJ <https://imagej.nih.gov/ij/index.html>`_ is an image processing and
+`ImageJ <https://imagej.net/>`_ is an image processing and
 analysis application written in Java, widely used in the life sciences
 fields, with an extensible plugin infrastructure. You can use
 Bio-Formats as a plugin for ImageJ to read and write images in the
--- a/sphinx/users/imagej/index.rst
+++ b/sphinx/users/imagej/index.rst
@@ -1,7 +1,7 @@
 ImageJ overview
 ===============
 
-`ImageJ <https://imagej.net/>`_ is an image processing and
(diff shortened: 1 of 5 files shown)


sphinx/users/imagej/installing.rst
```diff
@@ -10,7 +10,7 @@ Installing Bio-Formats in ImageJ
     <http://help.openmicroscopy.org/imagej.html>`_ useful
     for getting you started with both plugins at the same time.
 
-Once you `download <https://imagej.nih.gov/ij/download.html>`__ and
+Once you `download <https://imagej.net/downloads>`__ and
```
melissalinkert, line 13:
```suggestion
Once you `download <https://imagej.net/ij/download.html>`__ and
```
`imagej.net/ij/download.html` is the equivalent of the previously-referenced download page. imagej.net/downloads references Fiji as well, so may be confusing since the install instructions here are known to be incompatible with Fiji.

sphinx/users/imagej/index.rst
```diff
@@ -1,7 +1,7 @@
 ImageJ overview
 ===============
 
-`ImageJ <https://imagej.nih.gov/ij/index.html>`_ is an image processing and
+`ImageJ <https://imagej.net/>`_ is an image processing and
```
melissalinkert, line 4:
```suggestion
`ImageJ <https://imagej.net/ij/index.html>`_ is an image processing and
```
Point to the equivalent of the previously-referenced ImageJ main page, not the one that references Fiji/ImageJ2 since instructions here are not compatible with Fiji.

melissalinkert (commented):
I don't feel too strongly about the couple of suggestions added, so if there's a case for leaving as-is that's fine too. Otherwise looks good.

melissalinkert (approved):

PR #380, opened 2024-07-08
Add version history for 7.3.1

dgault:
Adding version history for the release of 7.3.1

For a full list of included PR's see https://github.com/ome/bioformats/milestone/101?closed=1


sbesson (changes requested):
Assuming https://github.com/ome/bioformats/pull/4193 gets included in the upcoming patch release, the history will need to be updated to capture this change as well

sbesson (approved):

melissalinkert (approved):

PR #388, opened 2024-07-30
Update 7.3.1 note on warnings to clarify potential broader impact

melissalinkert:
Following today's OME meeting and separate discussion with @chris-allan, this change updates the release notes in an attempt to clarify the potential impact of using `DataTools.parseDouble` as part of the warnings cleanup in https://github.com/ome/bioformats/pull/4182 and https://github.com/ome/bioformats/pull/4177.

### DIFF
--- a/sphinx/about/whats-new.rst
+++ b/sphinx/about/whats-new.rst
@@ -21,6 +21,13 @@ File format fixes and improvements:
 Bio-Formats improvements:
 
 * fixed warnings across all components for Java 9+ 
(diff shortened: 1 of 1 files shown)


chris-allan (approved):
At least to me, this feels sufficient. We do bug fixes all the time in point releases, this just happens to be a very big bug fix as well as doubling as Java 9 compatibility. It's good to have it called out from both a technical and good will perspective.

dgault (approved):

PR #392, opened 2024-09-11
Add documentation for showinf batch processing via stdin.

tstoeter:
Documentation for batch processing using showinf as introduced by ome/bioformats/pull/4200 and ome/bioformats/pull/4233.


melissalinkert (approved):
Thanks, @tstoeter. https://bio-formats--392.org.readthedocs.build/en/392/users/comlinetools/display.html#cmdoption-showinf-0 looks fine to me, if @sbesson is also happy with this change.

sbesson (approved):

PR #393, opened 2024-09-18
Use https instead of http

jburel:
Link to outbackphoto is sporadically breaking the daily build


sbesson (approved):
Assuming this is validated by the nightly CI builds

PR #394, opened 2024-09-26
Add sas.upenn.edu to ignore list linkchecker

pwalczysko:
Trying to fix failing build at https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/

cc @melissalinkert

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -390,5 +390,6 @@ linkcheck_ignore = ['https://imspector.mpibpc.mpg.de',
     r'http://openslide.cs.cmu.edu*', # NameResolutionError
     r'https://openslide.cs.cmu.edu*', # NameResolutionError
     r'https://tronche.com/computer-graphics/*', # NameResolutionError
-    r'https://www.sas.upenn.edu/*'  # Invalid SSL certificate
+    r'https://www.sas.upenn.edu/*',  # Invalid SSL certificate
+    r'http://www.bio-rad.com/*' # 403 Client Error: Forbidden for url
 ]
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -388,5 +388,9 @@ linkcheck_ignore = ['https://imspector.mpibpc.mpg.de',
     r'https://www.mayo.edu/research/core-resources/biomedical-imaging-resource-core/overview', # 403 Client Error: Forbidden
     r'https://trac.openmicroscopy.org*',
     r'http://openslide.cs.cmu.edu*', # NameResolutionError
-    r'https://openslide.cs.cmu.edu*' # NameResolutionError
+    r'https://openslide.cs.cmu.edu*', # NameResolutionError
+    r'https://sourceforge.net/projects/libmng*', # 403 Client Error: Forbidden
+    r'https://www.bio-rad.com*', # 403 Client Error: Forbidden
+    r'https://www.excelitas.com/*', # 403 Client Error: Forbidden
+
 ]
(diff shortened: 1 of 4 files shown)


melissalinkert (commented):
Leaving this open for a few days to fix the build is fine with me. The certificate only expired yesterday, so I would think there is a strong possibility it gets fixed upstream in that time.

If this is still a problem on Monday, though, I would propose to instead remove the link from `src/main/resources/format-pages.txt`, as ignoring the invalid certificate does not provide anything useful:

```
$ curl -k https://www.sas.upenn.edu/~heiney/html-physics/datasqueeze/
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>500 Proxy Error</title>
</head><body>
<h1>Proxy Error</h1>
The proxy server could not handle the request <em><a href="/~heiney/html-physics/datasqueeze/">GET&nbsp;/~heiney/html-physics/datasqueeze/</a></em>.<p>
Reason: <strong>Error during SSL Handshake with remote server</strong></p><p />
</body></html>
```

sphinx/conf.py
```diff
@@ -389,5 +389,6 @@ def getValue(value):
     r'https://trac.openmicroscopy.org*',
     r'http://openslide.cs.cmu.edu*', # NameResolutionError
     r'https://openslide.cs.cmu.edu*', # NameResolutionError
-    r'https://tronche.com/computer-graphics/*' # NameResolutionError
+    r'https://tronche.com/computer-graphics/*', # NameResolutionError
+    r'http://www.bio-rad.com/*' # 403 Client Error: Forbidden for url
```
jburel, line 393: Could you try to use https instead of excluding?

sphinx/conf.py
```diff
@@ -389,5 +389,6 @@ def getValue(value):
     r'https://trac.openmicroscopy.org*',
     r'http://openslide.cs.cmu.edu*', # NameResolutionError
     r'https://openslide.cs.cmu.edu*', # NameResolutionError
-    r'https://tronche.com/computer-graphics/*' # NameResolutionError
+    r'https://tronche.com/computer-graphics/*', # NameResolutionError
```
jburel, line 392:
```
Wednesday 24 July 2024 at 01:00:00
Expires On	Friday 15 August 2025 at 00:59:59
```
Certificate is now valid

pwalczysko: fixed in https://github.com/ome/bio-formats-documentation/pull/394/commits/a3ef84ee89ed69228628ee1232a41bb80df45a72

sphinx/conf.py
pwalczysko: fixed in https://github.com/ome/bio-formats-documentation/pull/394/commits/a3ef84ee89ed69228628ee1232a41bb80df45a72

pwalczysko:
This PR is not making it to the Jenkins build because of https://github.com/ome/bio-formats-documentation/pull/395#issuecomment-2406959482 - the sphinx build is broken for all PRs in this repo

pwalczysko:
I think the sphinx build problem is https://github.com/python/cpython/issues/104818 - we are using Sphinx v5.3.0 with Python 3.13 which does not work because Python 3.13 does not have the ``imghdr`` module.

[They propose 2 ways](https://github.com/python/cpython/issues/104818) of fixing as far as I understand:

- Sphinx 6.2.0 or
- Python 3.12 or lower

sbesson:
You're correct, the issue is that the versions of Python and Sphinx used in the current build are incompatible. Either we need a Python downgrade or a Sphinx upgrade.

Unfortunately, how the versions are ultimately specified and picked is fairly opaque and hidden between multiple layers (`.readthedocs.yml` -> `mambaforge` -> `environment.yml`). #396 is an initial attempt at making one of the two changes above but without success so far

pwalczysko:
rebased this on top of master to fix build as https://github.com/ome/bio-formats-documentation/pull/396 is merged. Thank you @sbesson

melissalinkert (approved):

PR #398, opened 2024-10-23
replace http by https

jburel:
Replace http by https


sbesson (approved):
Note the CI linkchecks are broken with connection timeouts but this is related to internal UoD firewall issues which are known and need to be fixed independently of this PR

PR #402, opened 2024-12-13
Fix link

pwalczysko:
Fixing build.

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -395,6 +395,6 @@ linkcheck_ignore = ['https://imspector.mpibpc.mpg.de',
     r'https://www.sas.upenn.edu/*', # certificate issuer cannot be verified
     r'https://www.microsoft.com/en-us/download/details.aspx?id=48145', # 403 Client Error: Forbidden
     r'https://www.agilent.com/home', # 403 Client Error: Forbidden
-    r'https://www.mathworks.com/*', # 403 Client Error: Forbidden
-    r'https://uk.mathworks.com' # 403 Client Error: Forbidden
+    r'https://mathworks.com/*', # 403 Client Error: Forbidden
+    r'https://uk.mathworks.com/*' # 403 Client Error: Forbidden
 ]
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -395,6 +395,6 @@ linkcheck_ignore = ['https://imspector.mpibpc.mpg.de',
     r'https://www.sas.upenn.edu/*', # certificate issuer cannot be verified
     r'https://www.microsoft.com/en-us/download/details.aspx?id=48145', # 403 Client Error: Forbidden
     r'https://www.agilent.com/home', # 403 Client Error: Forbidden
-    r'https://www.mathworks.com', # 403 Client Error: Forbidden
+    r'https://www.mathworks.com/*', # 403 Client Error: Forbidden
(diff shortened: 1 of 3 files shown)


sphinx/conf.py
```diff
@@ -394,5 +394,7 @@ def getValue(value):
     r'https://www.excelitas.com/*', # 403 Client Error: Forbidden
     r'https://www.sas.upenn.edu/*', # certificate issuer cannot be verified
     r'https://www.microsoft.com/en-us/download/details.aspx?id=48145', # 403 Client Error: Forbidden
-    r'https://www.agilent.com/home' # 403 Client Error: Forbidden
+    r'https://www.agilent.com/home', # 403 Client Error: Forbidden
+    r'https://www.mathworks.com', # 403 Client Error: Forbidden
```
melissalinkert, line 398: 403 errors still appear in https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/287/console. Looking at other examples above, it might be that `https://www.mathworks.com/*` needs to be used instead?

sphinx/conf.py
```diff
@@ -394,5 +394,7 @@ def getValue(value):
     r'https://www.excelitas.com/*', # 403 Client Error: Forbidden
     r'https://www.sas.upenn.edu/*', # certificate issuer cannot be verified
     r'https://www.microsoft.com/en-us/download/details.aspx?id=48145', # 403 Client Error: Forbidden
-    r'https://www.agilent.com/home' # 403 Client Error: Forbidden
+    r'https://www.agilent.com/home', # 403 Client Error: Forbidden
+    r'https://www.mathworks.com/*', # 403 Client Error: Forbidden
```
melissalinkert, line 398: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/288/ still shows errors for 6 `https://mathworks.com/...` links. Probably `r'https://mathworks.com/*'` needs to be added separately, or all links need to be unified to `https://www.mathworks.com/...` instead of a mix of that and `https://mathworks.com/...`.

pwalczysko: > still shows errors for 6 https://mathworks.com/... links

Do not see that. Instead, imho the build shows errors for ``https://www.mathworks.com`` - illogical, but true imho. Somehow the listing of ``'https://www.mathworks.com/*'`` did not bite in.

pwalczysko: I will try to fix it doing the exact opposite of your suggestion @melissalinkert -> get rid of all www.math... in favor of https:///mathworks.com

pwalczysko: Sorry, 

> Do not see that. Instead, imho the build shows errors for https://www.mathworks.com - illogical, but true imho. > Somehow the listing of 'https://www.mathworks.com/*' did not bite in.

Thta ^^^ statement of mine is impresice as well. The original prob is the ``https://mathworks...`` but the error is produced as if for ``https://www.mathworks...`` (after a redirect ?)

Anyway, I think my fix https://github.com/ome/bio-formats-documentation/pull/402/commits/3b318d08d304731f0157db2868bd4b9cbe970e72 should get a fair chance to solve it :)

pwalczysko:
@melissalinkert sorry, one more fix https://github.com/ome/bio-formats-documentation/pull/402/commits/546c80c3ebf4592c7f1c35113bb838b7da62f0e1 which should put it in a really good position now to be right :)

Issue #247, opened 2022-05-27
Third-party Bio-Formats extension


sbesson:
Potential candidate for adding to the list of third-party tools consuming Bio-Formats https://docs.openmicroscopy.org/bio-formats/latest/users/index.html

- R - https://github.com/aoles/RBioFormats
- Wolfram (incl. Mathematica) - https://github.com/WolframResearch/BioFormatsLink

sbesson:
A few thoughts related to this issue and motivated by the discussions in #84.

The [User information](https://bio-formats.readthedocs.io/en/latest/users/index.html) includes a lot of content about third-party integrations. Most of it has been inherited from a previous iteration of the Bio-Formats documentation over a decade ago which was maintained as a wiki. It is fair to say we haven't significantly expanded this section or even adjusted the workflow. As we are talking about additions/modification/deletions and there are growing discussions about the policy for community contributions, this feels fairly timely

There are a few issues with the way third-party tools are maintained in this documentation. Probably most of these stem from the fact that these integrations are neither developed nor tested by the Bio-Formats maintainer team.

- as noted in #83, several of these integrations might be in a stale state because the integration or even the application itself is no longer developed, supported or maintained
- the content of these individual pages usually includes a description of the application as well as installation instructions for the integration and is hard to maintain

Moving forward, my inclination would be to extract this third-party tooling content from the reference Bio-Formats documentation as follows:

- define the set of user facing tool that are officially maintained, tested and supported. Currently, I would define these as the command-line utilities, the Fiji/ImageJ plugin and the MATLAB toolbox
- extract all the content of the user documentation that does not meet this criterion to a new location to be determined e.g. a GitHub repository and/or a website registering third-party Bio-Formats integrations. In its simplest form, this could start as a single Markdown table with linking to other pages giving additional content for the given applications
- define the rules for managing the content of this third-party tools registry. As per the above, I would vote for not puttig this responsibility to Bio-Formats maintainers. If this content lives under the OME organization, probably a minimal amount of review should happen. Or it could live under a separate organization. 
- remove these items from the reference documentation and include a link to these third-party tools

There have been a few attempts at organizing community documentation  in the past e.g. https://github.com/ome/community-docs. In a similar vein, for OMERO, several third-party integrations have been migrated to https://omero-guides.readthedocs.io/. We should definitely build on the successes and failures of these experiences and ideally find a solution that aligns with the management of the OME training materials /cc @jburel @pwalczysko

[...]

PR #409, opened 2025-02-10
Link to Farsight GitHub instead of wiki

melissalinkert:
At the time of this commit, the wiki has been entirely removed, and `farsight-toolkit.org` no longer refers to the Farsight project.

The [linkcheck](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/338/) build has been failing for a few days, so this should hopefully fix the build.

I'm of multiple minds on the best approach here - on one hand, it's possible the wiki is just down for maintenance, and will come back at some point or be moved somewhere else. On the other hand, no updates to the GitHub repo in over a decade and no other usable web presence suggests that Farsight should be removed entirely from Bio-Formats documentation. I opted for just changing links to get the build passing, but happy to hear other thoughts. This ties in a bit with the discussion of #247, @sbesson / @joshmoore / @jburel.

### DIFF
--- a/sphinx/users/farsight/index.rst
+++ b/sphinx/users/farsight/index.rst
@@ -1,7 +1,7 @@
 FARSIGHT
 ========
 
-`FARSIGHT <http://farsight-toolkit.ee.uh.edu/wiki/Main_Page>`_ is a collection of
+`FARSIGHT <https://github.com/RoysamLab/Farsight-toolkit>`_ is a collection of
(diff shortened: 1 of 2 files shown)


joshmoore:
Seems like a good strategy considering:
 * No entry in the wayback machine!
 * The GH linked website (`https://www.farsight-toolkit.org/`) is *NOT* to be trusted.

sbesson (approved):

PR #410, opened 2025-02-13
Add page for Hamamatsu DCIMG

melissalinkert:
See https://github.com/ome/bioformats/pull/4255

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -808,6 +808,19 @@ utilityRating = Fair
 reader = NAFReader
 mif = true
 
+[Hamamatsu DCIMG]
+extensions = .dcimg
+owner = `Hamamatsu <https://www.hamamatsu.com/>`_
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Screenshot 2025-02-13 at 18 33 48

Should we get this is and tag this repository to update https://bio-formats.readthedocs.io/ ?

melissalinkert:
Only concern with tagging is that doing so outside of the normal release cycle might cause confusion? Maybe that's incentive to get an 8.1.1 out sooner rather than later?

sbesson:
I don't think we have used this strategy often for Bio-Formats but I recall one of the intended goals while moving to readthedocs was to be able to quickly update the documentation for a given version of the software. For the OMERO documentation, the convention was to push a `vx.y.z-b` tag with the suffix being incremented every time - see https://github.com/ome/omero-documentation/tags.

PR #414, opened 2025-03-03
Delete ionpath, ignore princeton

pwalczysko:
1. Delete ionpath links, as ``Company no longer exist`` message is on https://www.ionpath.com/
2. Ignore ``...princeton....`` as link works, but probably linkchecker is not coming through

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -396,5 +396,6 @@ linkcheck_ignore = ['https://imspector.mpibpc.mpg.de',
     r'https://www.microsoft.com/en-us/download/details.aspx?id=48145', # 403 Client Error: Forbidden
     r'https://www.agilent.com/home', # 403 Client Error: Forbidden
     r'https://mathworks.com/.*', # 403 Client Error: Forbidden
-    r'https://uk.mathworks.com/.*' # 403 Client Error: Forbidden
+    r'https://uk.mathworks.com/.*', # 403 Client Error: Forbidden
+    r'https://www.princetoninstruments.com/.*' # 403 Client Error: Forbidden
 ]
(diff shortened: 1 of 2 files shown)


sphinx/conf.py
```diff
@@ -396,5 +396,6 @@ def getValue(value):
     r'https://www.microsoft.com/en-us/download/details.aspx?id=48145', # 403 Client Error: Forbidden
     r'https://www.agilent.com/home', # 403 Client Error: Forbidden
     r'https://mathworks.com/.*', # 403 Client Error: Forbidden
-    r'https://uk.mathworks.com/.*' # 403 Client Error: Forbidden
+    r'https://uk.mathworks.com/.*', # 403 Client Error: Forbidden
+    r'https://www.princetoninstruments.com/.*' # 403 Client Error: Forbidden
```
melissalinkert, line 400: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/361/ still flags this line:

https://github.com/ome/bio-formats-documentation/blob/52c9abb371c950e81fc1cdaf19d09953ae88a99b/src/main/resources/format-pages.txt#L2230

as broken. I think that's because the link has no trailing slash. Either a trailing slash should be added, or the ignore should be `r'https://www.princetoninstruments.com*'` - I don't have a preference either way.

pwalczysko: Thank you for figuring that out. Fixed in https://github.com/ome/bio-formats-documentation/pull/414/commits/fa049af5a9ea7bf4697b2c9943be6d242a979e76.

There is a mixture of links with and without trailing slash in the https://github.com/ome/bio-formats-documentation/blob/52c9abb371c950e81fc1cdaf19d09953ae88a99b/src/main/resources/format-pages.txt. But ``with slash`` slightly wins.

melissalinkert (approved):
https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/364/

PR #415, opened 2025-03-03
Make sure all public data is referenced from format pages

melissalinkert:
Each of the folders under https://downloads.openmicroscopy.org/images/ is now referenced from a format page, with the exception of:

- https://downloads.openmicroscopy.org/images/HCS/ (appears to be symlinks, not additional data)
- https://downloads.openmicroscopy.org/images/gateway_tests/
- https://downloads.openmicroscopy.org/images/u-track/

The `MetaXpress` entry has also been updated to reference the correct reader (there is a separate Cellworx entry).

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -343,6 +343,7 @@ weHave = * an SDT specification document (from 2008 April, in PDF) \n
 * an SDT specification document (from 2006 June, in PDF) \n
 * Becker & Hickl's `SPCImage <https://www.becker-hickl.com/products/spcimage/>`_ software \n
 * a large number of SDT datasets \n
+* `public sample images <https://downloads.openmicroscopy.org/images/SDT/>`__ \n
 * the ability to produce new datasets
 pixelsRating = Very good
 metadataRating = Good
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Changes looks good. Like other documentation PRs, waiting to see whether the CI builds are passing before merging this.

The three folders you flagged are pre-date the pattern of 1 top-level folder == 1 file format/reader introduced since:
- `HCS` used to contain the public HCS samples until this content got migrated into the proper top-level HCS folders
- `u-track` / `gateway_tests` contain specific datasets used for testing mid 2010s

Thes folders files were previously hidden from the directory listing either via `.htaccess` or via configuration - see https://web.archive.org/web/20240809123644/https://downloads.openmicroscopy.org/images/ for a previous snapshot. This configuration might have been lost as part of the migration of https://downloads.openmicroscopy.org/ /cc @jburel @pwalczysko It's not really a big issue, possibly something to capture to reduce the confusion. As discussed at the weekly Formats meeting, the two files `OMERO.tar.gz` / `omero_db_searchengine.zip` introduced <12 months ago are also unrelated. If there was a mechanism to hide them without breaking the URL,  this might be another way to resolve the issue.

PR #416, opened 2025-03-03
Fix two more broken links

melissalinkert:
Both flagged by https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/360/. Together with #414, this is expected to fix the linkcheck build.

PR #418, opened 2025-03-14
Update MATLAB documentation for colormap return value

sbesson:
Documents the changes from https://github.com/ome/bioformats/pull/4291 and scheduled for release in the next minor release of Bio-Formats

### DIFF
--- a/sphinx/developers/matlab-dev.rst
+++ b/sphinx/developers/matlab-dev.rst
@@ -105,13 +105,19 @@ series in the dataset. If ``s`` is the series index between 1 and ``n``:
 -  The ``data{s, 2}`` element contains original metadata key/value pairs that
    apply to the ``s``-th series.
 
--  The ``data{s, 3}`` element contains color lookup tables for each plane in
-   the ``s``-th series.
+-  The ``data{s, 3}`` element is an ``m``-by-1 cell array, , where ``m`` is the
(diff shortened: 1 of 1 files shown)


sphinx/developers/matlab-dev.rst
```diff
@@ -105,14 +105,20 @@ series in the dataset. If ``s`` is the series index between 1 and ``n``:
 -  The ``data{s, 2}`` element contains original metadata key/value pairs that
    apply to the ``s``-th series.
 
--  The ``data{s, 3}`` element contains color lookup tables for each plane in
-   the ``s``-th series.
+-  The ``data{s, 3}`` element is an ``m``-by-1 cell array, , where ``m`` is the
```
melissalinkert, line 108:
```suggestion
-  The ``data{s, 3}`` element is an ``m``-by-1 cell array, where ``m`` is the
```

melissalinkert:
Minor typo aside, this looks good. My only concern with merging right away is that it means we probably shouldn't do any intermediate tags before 8.2.0 (e.g. for linkcheck fixes). It might be worth getting #414 / #416 merged and tagged before this is merged since the linkcheck job has stabilized (cc @pwalczysko).

sbesson:
:+1: agreed with holding off on merging this until we are closer to a Bio-Formats 8.2.0 release.

melissalinkert (approved):

PR #428, opened 2025-07-04
Fix some link anchors

dominikl:
Bioformats linkcheck job keeps failing since a while with these four links broken. See https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/498/ .


sbesson (changes requested):
Sorry for having missed those, I think this is a consequence of the fact the latest `ome-common/ome-model` releases have been compiled with JDK 11 (with the `--release 8` java flag) and the Javadoc format has slightly changed.

For all the links, I think the change is to replace `-` by `(`, `,` or `)` depending on the case e.g.

https://javadoc.io/doc/org.openmicroscopy/ome-common/6.0.25/loci/common/Location.html#mapFile-java.lang.String-loci.common.IRandomAccess-

became

https://javadoc.io/doc/org.openmicroscopy/ome-common/6.0.26/loci/common/Location.html#mapFile(java.lang.String,loci.common.IRandomAccess)

melissalinkert:
Closing, as #429 more thoroughly addresses this issue and preserves the correct signatures.

PR #434, opened 2025-08-05
Add format page for Zeiss XRM format

sbesson:
Following the release of the 8.3.0 components, the `ZeissXRMReader` metadata page should be generated and cross-linked.
Up for review as always are the format ratings as well as the various informations about the reference software and test datasets


melissalinkert (changes requested):
Only other thing apart from couple of minor comments is that I couldn't review the built page as the RTD build link goes to https://bio-formats--434_a.org.readthedocs.build/en/434_a/ which currently just returns 400.

melissalinkert:
Looks good, just closing/re-opening to try to force a build.

melissalinkert (approved):

sbesson:
Now deployed at https://bio-formats.readthedocs.io/en/stable/formats/zeiss-xrm.html

PR #438, opened 2025-09-19
Fix netcdf-java links

will-moore:
update broken links


will-moore:
Thanks. The 2nd of those links looked best. Updated

PR #440, opened 2025-09-29
Exclude https://www.mail-archive.com/tiff from linkcheck

dominikl:
Exclude https://www.mail-archive.com/tiff@lists.osgeo.org/msg00421.html from linkcheck, as it often times out.

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -398,5 +398,5 @@ linkcheck_ignore = ['https://imspector.mpibpc.mpg.de',
     r'https://uk.mathworks.com/.*', # 403 Client Error: Forbidden
     r'https://www.princetoninstruments.com/.*', # 403 Client Error: Forbidden
     r'https://www.merckmillipore.com*', # 403 Client Error: Forbidden
-    r'https://www.mail-archive.com/.*' # Connection to www.mail-archive.com often times out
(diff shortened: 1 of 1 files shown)


pwalczysko (changes requested):
Atm does not work as expected, see above.

melissalinkert:
> I think we could try to exclude `https://www.mail-archive.com*` (i.e. not slash-dot-star but just star). What do you think @melissalinkert please ?

That approach makes sense to me.

melissalinkert:
If I run `mvn test -DskipSphinxTests=false -Dome-model.uri=/bio-formats-build/ome-model/docs/sphinx/target/sphinx/html` locally with this change, then the results in `target/sphinx/linkcheck/output.json` show that the link is correctly ignored:

```
$ grep mail-archive output.json 
{"filename": "about/whats-new.rst", "lineno": 230, "status": "ignored", "code": 0, "uri": "https://www.mail-archive.com/tiff@lists.osgeo.org/msg00421.html", "info": ""}
```

It's not clear to me why https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/610/ and preceding builds continue to fail. If I remember correctly, we had similar issues with #430, so perhaps a similar fix on the CI side is needed?

PR #453, opened 2026-01-08
Fix links

pwalczysko:
2 links breaking the build repeatedly.

1. in https://github.com/ome/bio-formats-documentation/commit/3de2bf5574c02194e75cf726bdcac859347ccf05 ignoring the pdf file from linkchecker as the link itself is working fine
2. in https://github.com/ome/bio-formats-documentation/commit/07be57f1eeb26fd91e26205c0f7aaf63791f91ee suggesting a different link, as the whole www.nikondigital.org seems to be broken for a week now. The question is @melissalinkert whether we can/should point to https://filesamples.com for sample files ?

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -1736,7 +1736,6 @@ extensions = .nef, .tif
 developer = `Nikon <http://www.nikon.com/>`_
 bsd = no
 samples = `neffile1.zip <https://www.outbackphoto.com/workshop/NEF_conversion/neffile1.zip>`_ \n
-`Sample NEF images <https://filesamples.com/formats/nef>`_
 weHave = * a NEF specification document (v0.1, from 2003, in PDF) \n
 * several NEF datasets
 pixelsRating = Very good


src/main/resources/format-pages.txt
```diff
@@ -1736,7 +1736,7 @@ extensions = .nef, .tif
 developer = `Nikon <http://www.nikon.com/>`_
 bsd = no
 samples = `neffile1.zip <https://www.outbackphoto.com/workshop/NEF_conversion/neffile1.zip>`_ \n
-`Sample NEF images <http://www.nikondigital.org/articles/library/nikon_d2x_first_impressions.htm>`_
+`Sample NEF images <https://filesamples.com/formats/nef>`_
```
melissalinkert: I'd be inclined to just remove this line for now, especially as we won't have tested the linked files (so they may not work). https://github.com/ome/bioformats/issues/4389 has a variety of public samples linked, which we can add here after that issue is fixed.

pwalczysko: fixed in https://github.com/ome/bio-formats-documentation/pull/453/commits/ab5770f1de18dc342625a9dfa947b73c8dd5e3e3

PR #459, opened 2026-03-03
Replace velocity dependency by velocity-engine-core

sbesson:
There has been no new releases of the org.apache.velocity:velocity dependency since version 1.7. It looks like the templating engine is now released as org.apache.velocity:velocity-engine-core.


melissalinkert (approved):
The supported formats table looks fine, and spot checking a variety of format pages also shows no issues.

PR #468, opened 2026-05-11
Require Java 11 at runtime, build on Java 21

melissalinkert:
Prerequisite to merging #456. See https://www.openmicroscopy.org/2026/03/24/end-of-java-8-support.html.


sbesson (approved):
Given this a downstream repository solely used for documentation purposes, I would tend to say there are no implications to merging this immediately?

PR #232, opened 2022-02-17
TIFF standard: Update link

dgault:
Link to the TIFF standard has been broken on the adobe site for a number of days. See posts [Adobe & TIFF spec](https://community.adobe.com/t5/using-the-community-discussions/broken-link-to-open-standards-on-adobe-io/m-p/12753322) and ​​[second post](https://experienceleaguecommunities.adobe.com/t5/adobe-i-o-cloud-extensibility/broken-link-to-open-standards-on-adobe-io/td-p/441124).

PR #471, opened 2026-05-25
Replace faqs.org links

dominikl:
See title. Because they've moved http://www.faqs.org behind a captcha test.

### DIFF
--- a/src/main/resources/format-pages.txt
+++ b/src/main/resources/format-pages.txt
@@ -2204,7 +2204,7 @@ utilityRating = Poor
 reader = PictReader
 \n
 .. seealso:: \n
-  `PICT technical overview <http://www.faqs.org/faqs/graphics/fileformats-faq/part3/section-107.html>`_ \n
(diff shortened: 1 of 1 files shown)


src/main/resources/format-pages.txt
```diff
@@ -2727,7 +2727,7 @@ reader = BMPReader
 notes = Compressed BMP files are currently not supported. \n
 \n
 .. seealso:: \n
-  `Technical Overview <http://www.faqs.org/faqs/graphics/fileformats-faq/part3/section-18.html>`_ \n
+  `Technical Overview <https://en.wikipedia.org/wiki/BMP_file_format>`_ \n
```
sbesson: An alternative would be the Library of Congress reference page - https://www.loc.gov/preservation/digital/formats/fdd/fdd000189.shtml. This is similar to what we did with TIFF - see https://github.com/ome/bio-formats-documentation/pull/232

pwalczysko (approved):
Thank you, lgtm

